PR #1, opened 2019-03-07
Adapt istio-installer after testing

c0d1ngm0nk3y:
Please have a look at the issues we faced during using the istio-installer.


c0d1ngm0nk3y:
@costinm Could you have a look? Does it make sense?

PR #2, opened 2019-03-12
Adapt documentation

c0d1ngm0nk3y:
Document findings after trying out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ it is possible to use other tools that create the expected certificates for Isti
 iop istio-system istio-system-security $IBASE/istio-system-security
 ```
 
-Important options: the 'dnsCerts' list allows associating DNS certs with specific service accounts.
+**Important options**: the 'dnsCerts' list allows associating DNS certs with specific service accounts.
 This should be used if you plan to use Galley or Sidecar injector in different namespaces.
 By default it supports "istio-control", "istio-master" namespaces used in the examples.
 
@@ -195,7 +195,7 @@ and it is recommended to have Pilot running in each region and in multiple avail
                --set global.istioNamespace=istio-master \
                --set global.configNamespace=istio-master \
                --set global.telemetryNamespace=istio-telemetry-master \
-               --set global.policyNamespace=istio-policy-maste
+               --set global.policyNamespace=istio-policy-master
 
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -241,9 +241,19 @@ Pilot instance in the ingress namespace. This will be fixed in future releases.
 Note that running a dedicated Pilot for ingress/gateways is supported and recommended for very large sites, 
 but in the case of K8S ingress it is currently required.
 
+```bash
+    iop istio-ingress istio-ingress $IBASE/istio-installer/gateways/istio-ingress  --set global.istioNamespace=istio-master
```
c0d1ngm0nk3y: @costinm This is the case you described above, right? So it should not work? Actually, it didn't :( We would have to unset istioNamespace and install another pilot?

README.md
```diff
@@ -144,7 +144,7 @@ it is possible to use other tools that create the expected certificates for Isti
 iop istio-system istio-system-security $IBASE/istio-system-security
 ```
 
-Important options: the 'dnsCerts' list allows associating DNS certs with specific service accounts.
+**Important options**: the 'dnsCerts' list allows associating DNS certs with specific service accounts.
 This should be used if you plan to use Galley or Sidecar injector in different namespaces.
 By default it supports "istio-control", "istio-master" namespaces used in the examples.
```
c0d1ngm0nk3y: Would 'istio-control-master' not be the better name? It would be more consistent with the naming or are we on the wrong track?

README.md
```diff
@@ -241,9 +241,19 @@ Pilot instance in the ingress namespace. This will be fixed in future releases.
 Note that running a dedicated Pilot for ingress/gateways is supported and recommended for very large sites, 
 but in the case of K8S ingress it is currently required.
 
+```bash
```
c0d1ngm0nk3y: we added:
```
./bin/iop istio-ingress istio-ingress-pilot $IBASE/istio-control/istio-discovery --set global.istioNamespace=istio-control-master --set global.configNamespace=istio-control-master --set global.telemetryNamespace=istio-telemetry
```

to add another pilot, but the pilot won't start

```
Error receiving MCP resource: rpc error: code = Unavailable desc = no healthy upstream
```

What is the missing upstream in this case?

README.md
costinm: Sounds good - happy to merge a pull request ! 
Or maybe 'istio-control-canary' ? Since the purpose is to test the next version, master happens to be the branch but we may switch to 'master is stable'.

README.md
costinm: I think I need more logs - and some debugging. 
But I would suspect the MCP certificates. I only tested pilot in ingress namespace with galley off ( the goal
was to test the most minimal possible config )

README.md
costinm: Sorry :-), I can try it again. 

Try it first with istioNamespace=istio-control ( that's what I tested the most ). And logs/more info on what didn't work - usually it's some config that is not properly set.

PR #7, opened 2019-03-25
Add istio-version to globals.yaml.

jsPlachetka:
The version is needed to provide the correct flags to galley.

Co-authored-by: Axel Siebenborn <axel.siebenborn@sap.com>
Signed-off-by: Julia Plachetka <julia.plachetka@sap.com>

### DIFF
--- a/istio-control/istio-config/templates/deployment.yaml
+++ b/istio-control/istio-config/templates/deployment.yaml
@@ -36,7 +36,9 @@ spec:
           - containerPort: 15019
           command:
           - /usr/local/bin/galley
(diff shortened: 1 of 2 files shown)


istio-control/istio-config/templates/deployment.yaml
```diff
@@ -36,7 +36,9 @@ spec:
           - containerPort: 15019
           command:
           - /usr/local/bin/galley
+{{- if and (.Values.global.istioVersion) (semverCompare "<1.2" .Values.global.istioVersion) }}
```
costinm: Is this because they removed 'server' in master ? I think a better fix is to ask them to revert that change..
For now I'll merge this.

PR #19, opened 2019-04-02
Move MCP-address to mesh config (instead of using flag).

phil9909:
The `--mcpServerAddrs` flag was deprecated for a while now and removed from master this morning (https://github.com/istio/istio/commit/612c138738446644d0db747604353e83f441e166).
In order for the installer to work with master, the value should be provided via the meshConfig.
This is also compatible with istio-1.1.

### DIFF
--- a/istio-control/istio-discovery/templates/configmap.yaml
+++ b/istio-control/istio-discovery/templates/configmap.yaml
@@ -76,6 +76,14 @@ data:
     enableSdsTokenMount: true
     {{- end }}
 
+    {{- if .Values.useMCP }}
+    config_sources:
+    {{- if .Values.global.controlPlaneSecurityEnabled }}
+    - address: localhost:15019
+    {{- else }}
+    - address: istio-galley
+    {{- end }}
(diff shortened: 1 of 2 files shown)


istio-control/istio-discovery/templates/configmap.yaml
```diff
@@ -76,6 +76,14 @@ data:
     enableSdsTokenMount: true
     {{- end }}
 
+    {{- if .Values.useMCP }}
+    config_sources:
+    {{- if .Values.global.controlPlaneSecurityEnabled }}
+    - address: localhost:15019
+    {{- else }}
+    - address: istio-galley
```
costinm: Please check the envoy file used for 'controlPlaneSecurityEnabled' - the address should have the suffix, since galley may not run in the same namespace (or cluster) with pilot.

costinm: Technically it's not a regression - previous code also had it, so not blocking.

PR #22, opened 2019-04-09
Concourse ci update

c0d1ngm0nk3y:
Added update test to pipeline and checking bookinfo up that there is no downtime while updating.


.concourse/Dockerfile
```diff
@@ -11,7 +11,7 @@ FROM    golang:1.11
 
 # Also installed docker, docker-compose, dockerize, jq
 
-ARG     K8S_VER=v1.13.4
+ARG     K8S_VER=v1.14.0
```
costinm: FYI: on costinm/istio-installer I am trying to create a single "istio-builder" image set that can be used in all CI/CD systems, based on kindest/node ( i.e. the image used to run kind ). The image has built-in kubectl.

The goal would be to have  a set like istio-builder:k8s1.14, etc - so we can test multiple versions.

.concourse/bookinfo-watch.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+# This script runs the commands passed to it in parallel to a curl against the bookinfo application
```
costinm: If it doesn't have anything concourse-specific ( and it doesn't seem to have ), please move under test directory or bin.

bin/install.sh
```diff
@@ -30,80 +30,116 @@ function install_crds() {
 # Install citadel into namespace istio-system
 
 function install_system() {
-    kubectl delete namespace istio-system --wait --ignore-not-found
-    echo "Installing citadel.."
-    bin/iop istio-system istio-system-security $IBASE/security/citadel/ --set global.defaultResources.requests.cpu=0m
+    if [ "$METHOD" = Install ]; then
```
costinm: Perfect - just filed a bug about this :-)

bin/install.sh
```diff
@@ -30,80 +30,116 @@ function install_crds() {
...
-
-    kubectl get deployments -n istio-control
-    kubectl wait deployments istio-galley istio-pilot istio-sidecar-injector -n istio-control --for=condition=available --timeout=$WAIT_TIMEOUT
-    kubectl rollout status  deployment istio-galley -n istio-control --timeout=$WAIT_TIMEOUT 
-    kubectl rollout status  deployment istio-pilot  -n istio-control --timeout=$WAIT_TIMEOUT 
-    kubectl rollout status  deployment istio-sidecar-injector -n istio-control --timeout=$WAIT_TIMEOUT
-
-    kubectl get deployments -n istio-control
-    kubectl get pod -n istio-control 
+    kubectl delete namespace ${ISTIO_CONTROL_NS} --wait --ignore-not-found
```
costinm: Can you skip delete ? 

In all CI/CD systems the test is run in a clean cluster AFAIK.

gateways/istio-ingress/templates/deployment.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
     istio: ingressgateway
     release: {{ .Release.Name }}
 spec:
+  strategy:
```
costinm: Is this something we want in production ? 

Can you use a kustomize template ? That seems to be the consensus for changing k8s-specific things.

c0d1ngm0nk3y: We patch it anyway, so we don't need it. We will remove it.

But why would you this *NOT* in production? Otherwise, you would have a downtime during update, right?

loewenstein:
@costinm I was about to ask the same question that Ralf did... What would be a reason not to use rolling update in production?

gateways/istio-ingress/templates/deployment.yaml
costinm: We should absolutely do rolling update - but not with hardcoded parameters you have.

That's why I suggested kustomize template or values.yaml

costinm:
Sorry, I was not clear - rolling upgrade is great, but the params should not be hardcoded.

loewenstein:
@c0d1ngm0nk3y @jsPlachetka why do we remove watching?

@costinm isn't maxUnavailanle=0 a good default? If a deployment is scaled well, no matter if manually or with HPA it doesn't fall behind scale requirements even during update. I am a big fan of defaults, people should get what we would recommend without much configuration. If people really want different values, they can still override...
I know `kustomize` is bad with removing stuff, so if dev setups without rolling update is your concern, I'd still add it with a kustomization that is on by default.

gateways/istio-ingress/templates/deployment.yaml
c0d1ngm0nk3y: > We should absolutely do rolling update - but not with hardcoded parameters you have.
> 
> That's why I suggested kustomize template or values.yaml

I agree that this is something you ultimately want to have configurable. But a hardcoded rolling update would still be better than the default, right?

c0d1ngm0nk3y:
> @c0d1ngm0nk3y @jsPlachetka why do we remove watching?
> 

short answer: because it did not work.

long answer: There were 2 problems

* The way it was implemented (background tasks) did not work well with concourse. The job was very unstable because of some signals sent to the job.

* It does not work without downtime. At least the switching of the control planes is very likely to produce a failing curl since the bookinfo app uses Flask and Flask is not meant to be for productive usage. For example it does not drain existing connection when the pod is restarted.

c0d1ngm0nk3y:
I think we have addressed all comments and rebased it to master (conflicts). Please have a look.

PR #32, opened 2019-04-16
Delete old ns BEFORE restarting pods.

jkbschmid:
We switch istio-env to the new istio-control, first delete the old istio-control namespace before restarting pods. This prevents pods from using the old istio-control without us noticing.

Co-authored-by: Ralf Pannemans <ralf.pannemans@sap.com>

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -95,8 +95,8 @@ function switch_istio_control() {
         for ns in $(kubectl get namespaces --no-headers -l istio-env=${ISTIO_CONTROL_OLD} -o=custom-columns=NAME:.metadata.name); do        kubectl label namespaces default --overwrite istio-env=${ISTIO_CONTROL_NS}
             kubectl label namespaces ${ns} --overwrite istio-env=${ISTIO_CONTROL_NS}
         done
-        kubectl set env --all deployment --env="LAST_MANUAL_RESTART=$(date +%s)" --namespace=default
         kubectl delete namespace $ISTIO_CONTROL_OLD --wait --ignore-not-found
+        kubectl set env --all deployment --env="LAST_MANUAL_RESTART=$(date +%s)" --namespace=default
     fi
 }
 
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -92,11 +92,16 @@ function install_telemetry() {
 # Switch to other istio-control-namespace
 function switch_istio_control() {
     if [ "$ISTIO_CONTROL_NS" != "$ISTIO_CONTROL_OLD" ]; then
-        for ns in $(kubectl get namespaces --no-headers -l istio-env=${ISTIO_CONTROL_OLD} -o=custom-columns=NAME:.metadata.name); do        kubectl label namespaces default --overwrite istio-env=${ISTIO_CONTROL_NS}
+        ACTIVE_NAMESPACES=$(kubectl get namespaces --no-headers -l istio-env=${ISTIO_CONTROL_OLD} -o=custom-columns=NAME:.metadata.name)
+        for ns in $ACTIVE_NAMESPACES; do
             kubectl label namespaces ${ns} --overwrite istio-env=${ISTIO_CONTROL_NS}
         done
-        kubectl delete namespace $ISTIO_CONTROL_OLD --wait --ignore-not-found
-        kubectl set env --all deployment --env="LAST_MANUAL_RESTART=$(date +%s)" --namespace=default
+        if [ $REMOVE_OLD_CONTROL = true ]; then
+            kubectl delete namespace $ISTIO_CONTROL_OLD --wait --ignore-not-found
+        fi
+        for ns in $ACTIVE_NAMESPACES; do
+            kubectl set env --all deployment --env="LAST_MANUAL_RESTART=$(date +%s)" --namespace=$ns
+        done
     fi
 }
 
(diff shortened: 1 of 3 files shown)


jkbschmid:
@loewenstein Good point. In our test setup, we want to delete the old namespace so that we notice once pods keep using it instead of switching to the new one.

bin/install.sh
```diff
@@ -92,25 +92,34 @@ function install_telemetry() {
...
+            kubectl set env --all deployment --env="LAST_MANUAL_RESTART=$(date +%s)" --namespace=$ns
+        done
     fi
 }
 
 COMMAND="install_all"
 METHOD=Install
 ISTIO_CONTROL_OLD=$(kubectl get namespaces -o=jsonpath='{$.items[:1].metadata.labels.istio-env}' -l istio-env)
 ISTIO_CONTROL_OLD=${ISTIO_CONTROL_OLD:-istio-control}
+REMOVE_OLD_CONTROL=false
```
costinm: Please use ${REMOVE_OLD_CONTROL:-false}

bin/install.sh
```diff
@@ -92,25 +92,34 @@ function install_telemetry() {
...
 
 while [ $# -gt 0 ]
 do
     case "$1" in
         --update)
             METHOD=Update
             ;;
+        --remove-old-control)
```
costinm: I would prefer we use env variables consistently for options. Easier on the CI/CD and makefile.
The installer will probably move into istioctl anyways - but even there uncommon options should be set as env variables.

One option is for 'installer' to be packed as a docker container, with helm, istioctl, kubect - also easier to customize
various things via -e.

bin/test.sh
```diff
@@ -60,18 +60,15 @@ BOOKINFO_DEPLOYMENTS="details-v1 productpage-v1 ratings-v1 reviews-v1 reviews-v2
...
-    kubectl label ns bookinfo istio-env=istio-control --overwrite
-    #kubectl delete -f samples/bookinfo/platform/kube/bookinfo.yaml --ignore-not-found
-    #kubectl delete -f samples/bookinfo/networking/destination-rule-all-mtls.yaml --ignore-not-found
-    #kubectl delete -f samples/bookinfo/networking/bookinfo-gateway.yaml --ignore-not-found
+    ISTIO_CONTROL=$(kubectl get namespaces -o=jsonpath='{$.items[:1].metadata.labels.istio-env}' -l istio-env)
+    ISTIO_CONTROL=${ISTIO_CONTROL:-istio-control}
+
+    kubectl label ns bookinfo istio-env=${ISTIO_CONTROL} --overwrite
 
     kubectl -n bookinfo apply -f samples/bookinfo/platform/kube/bookinfo.yaml
```
costinm: Would it be possible to switch this to a fortio deployment ? Bookinfo is super heavy ( java, etc), we can leave it in post-submit / stability tests.

bin/test.sh
```diff
@@ -60,18 +60,15 @@ BOOKINFO_DEPLOYMENTS="details-v1 productpage-v1 ratings-v1 reviews-v1 reviews-v2
 
 if [ "$SKIP_SETUP" -ne 1 ]; then
     kubectl create ns bookinfo || /bin/true
-    kubectl label ns bookinfo istio-env=istio-control --overwrite
-    #kubectl delete -f samples/bookinfo/platform/kube/bookinfo.yaml --ignore-not-found
-    #kubectl delete -f samples/bookinfo/networking/destination-rule-all-mtls.yaml --ignore-not-found
-    #kubectl delete -f samples/bookinfo/networking/bookinfo-gateway.yaml --ignore-not-found
+    ISTIO_CONTROL=$(kubectl get namespaces -o=jsonpath='{$.items[:1].metadata.labels.istio-env}' -l istio-env)
```
costinm: A small comment on what this does ?

bin/install.sh
```diff
@@ -30,7 +30,7 @@ function install_crds() {
 # Cleanup all namespaces
 
 function cleanup() {
```
costinm: Why ? What's the use case ?

bin/test.sh
jkbschmid: Using something more lightweight than Bookinfo is a good idea. Fortio has the following problems, though:
* Fortio doesn't support graceful shutdowns, i.e. it can send incomplete responses during shutdown. This is a problem when we have to restart pods while switching to a new control plane.
* Fortio starts serving HTTP-requests before the server is fully configured. This could result in 404s during startup.

Should we try to fix these issues in fortio or should we write a small HTTP server which handles startup and shutdown correctly?
Our target is to prove that an istio update can be done in an HA setup.

bin/install.sh
jkbschmid: We always use the same cluster to repeatedly install and update istio. That's why we need a way to reset the cluster state.

costinm: We should separate test env scripts/configs from the 'install.sh' script. 
Maybe use the makefile to cleanup ?

Also - would it be possible to use the PR number or commit to name the environment ? 
That's the point of multi-environment, and most tests should be isolated (and if not we 
have a bug ). Same for fortio/bookinfo - they don't need to be installed in default ns, 
but some ns derived from the PR/commit.

costinm: BTW - few benefits are that you can use a cluster for multiple PRs, and we also verify that 
Isolation works and that indeed the new installer supports multiple versions of Istio at the 
same time, which is one of the main features. 

The namespaces can be cleaned at stop, or maybe with a cron job - for example we could
delete if the test is passing, and leave it around for 6 hours or so if it is not passing so
it can be debugged. I guess we could even create a ns-limited service account  in each
cluster so developers can easily debug (but without giving them too broad permission).

costinm:
We do need to test bookinfo anyways - but probably by porting the bookinfo test to the new installer. 
But this can be a post-submit or can be only in the 'stability' cluster.

And we do need to add the 'simple' test based on Fortio and some of the stability tests as well.
Of course - separate PRs.

PR #41, opened 2019-04-18
Consistent updates

jkbschmid:
We have restructured to run the following tests on our concourse pipeline:
* Install daily-1.1 images to test for compatibility between installer and daily 1.1
* Install release-1.1 images to test backwards compatibility with release 1.1
* Update from release 1.1 to daily-1.1 to verify updating works

Moreover:
* Move cleanup from install.sh to pipeline-yaml
* Auto-detect whether we update or install (remove all flags for install.sh)
* Bugfixes: 
** Redeploy bookinfo pods if present so they get current config
** Use retries for getting INGRESS_URL, since load balancer IP is added asynchronously.


bin/test.sh
```diff
@@ -75,19 +79,17 @@ if [ "$SKIP_SETUP" -ne 1 ]; then
...
-    export INGRESS_HOST=$(kubectl -n istio-ingress get service istio-ingressgateway -o jsonpath='{.spec.clusterIP}')
-fi
-export INGRESS_PORT=$(kubectl -n istio-ingress get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="http2")].port}')
-export SECURE_INGRESS_PORT=$(kubectl -n istio-ingress get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="https")].port}')
-export GATEWAY_URL=$INGRESS_HOST:$INGRESS_PORT
 set +e
 n=1
 while true
 do
+    export INGRESS_HOST=$(kubectl -n istio-ingress get service istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
```
loewenstein: Can we wait for a condition instead of retries?

bin/test.sh
```diff
@@ -59,11 +59,15 @@ cd $ISTIO_PATH
 BOOKINFO_DEPLOYMENTS="details-v1 productpage-v1 ratings-v1 reviews-v1 reviews-v2 reviews-v3"
 
 if [ "$SKIP_SETUP" -ne 1 ]; then
-    kubectl create ns bookinfo || /bin/true
+    kubectl create ns bookinfo || true
     # We use the first namespace with sidecar injection enabled to determine the control plane's namespace.
     ISTIO_CONTROL=$(kubectl get namespaces -o=jsonpath='{$.items[:1].metadata.labels.istio-env}' -l istio-env)
     ISTIO_CONTROL=${ISTIO_CONTROL:-istio-control}
 
+    kubectl -n bookinfo delete -f samples/bookinfo/platform/kube/bookinfo.yaml --ignore-not-found
```
costinm: Why ?

PR #42, opened 2019-04-18
Mtls default to permissive

costinm:
Since this is an 'a-la-carte' installer, we need to default to permissive mode, since 
components and features will be added by user.

With isolation and Sidecar - we should strongly recommend that users explicitly set
policy, rules, etc at namespace level instead of a global default.

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -184,7 +184,7 @@ global:
   mtls:
     # Default setting for service-to-service mtls. Can be set explicitly using
     # destination rules or service annotations.
-    enabled: true
+    enabled: false
 
   # ImagePullSecrets for all ServiceAccount, list of secrets in the same namespace
   # to use for pulling any images in pods that reference this ServiceAccount.


loewenstein:
@sdake I suppose there are situations where creating a CR during installation makes a lot of sense. And the ordering problem between CRD and CR can be solved.

costinm:
Both on and off installs some resources - off creates MeshPolicy with PERMISSIVE mode, on with STRICT.

I'm changing the tests to be explicit - and we should recommend users are explicit as well, i.e. when they deploy an app with Istio they should not rely on  the mesh default, but always include a Policy and Sidecar and DestinationRule. 

Like with Mixer Policy server, if users don't have any RBAC rules or policies defined - then PERMISSIVE
and STRICT are very similar, i.e. any workload in the mesh can connect to a service. With RBAC rule - 
non-mtls connections will be rejected since the identity of the caller can't be asserted.

loewenstein:
> I'm changing the tests to be explicit - and we should recommend users are explicit as well, i.e. when they deploy an app with Istio they should not rely on the mesh default, but always include a Policy and Sidecar and DestinationRule.

Not sure I agree. The more defaults we provide the less users have to do and think about..

PR #45, opened 2019-04-22
Only allow one replica of Citadel

lei-tang:
This PR is for the issue: https://github.com/istio/istio/issues/13383. The counterpart of this PR on istio/istio repo is at https://github.com/istio/istio/pull/13519.

### DIFF
--- a/security/citadel/templates/deployment.yaml
+++ b/security/citadel/templates/deployment.yaml
@@ -10,7 +10,7 @@ metadata:
     release: {{ .Release.Name }}
 
 spec:
-  replicas: {{ .Values.replicaCount }}
+  replicas: 1
   strategy:
     rollingUpdate:
       maxSurge: 1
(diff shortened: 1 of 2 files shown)


loewenstein:
@lei-tang can we make Citadel work with multiple replicas instead? With SDS Citadel gets on a more critical path, doesn't it? If a workload gets scheduled on a node the first time, the node agent needs the CSR done by Citadel in order to start serving traffic, right?

lei-tang:
In SDS-Citadel flow, Citadel may in theory have multiple replicas to serve CSRs from Node Agents. But overall multi-replica Citadel is still under development; currently for secret-mount-based Citadel when multiple Citadel are deployed, they have race condition. Per the quote from @jcetkov in https://github.com/istio/istio/issues/13383: "it might be appropriate to modify the helm charts not to allow setting the security replica count via variable, if such case is really not supported and results in hard to trace issues."

PR #50, opened 2019-04-23
don't use mtls for bookinfo destination rules in test.sh

c0d1ngm0nk3y:
As mtls was disabled in PR #42

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -93,6 +93,7 @@ function switch_istio_control() {
         done
         if [ $REMOVE_OLD_CONTROL = true ]; then
             kubectl delete namespace $ISTIO_CONTROL_OLD --wait --ignore-not-found
+            kubectl delete mutatingwebhookconfiguration istio-sidecar-injector-$ISTIO_CONTROL_OLD --wait --ignore-not-found
         fi
         for ns in $ACTIVE_NAMESPACES; do
             kubectl set env --all deployment --env="LAST_MANUAL_RESTART=$(date +%s)" --namespace=$ns
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -51,6 +51,21 @@ function install_control() {
     step "auto-injector.."
     bin/iop ${ISTIO_CONTROL_NS} istio-autoinject $IBASE/istio-control/istio-autoinject --set global.istioNamespace=${ISTIO_CONTROL_NS} --set global.istioNamespace=${ISTIO_CONTROL_NS} $RESOURCES_FLAGS
 
+    # Assure that webhook is deleted whenever sidecar-injector is deleted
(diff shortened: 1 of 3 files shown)


bin/install.sh
```diff
@@ -51,6 +51,21 @@ function install_control() {
     step "auto-injector.."
     bin/iop ${ISTIO_CONTROL_NS} istio-autoinject $IBASE/istio-control/istio-autoinject --set global.istioNamespace=${ISTIO_CONTROL_NS} --set global.istioNamespace=${ISTIO_CONTROL_NS} $RESOURCES_FLAGS
 
+    # Assure that webhook is deleted whenever sidecar-injector is deleted
```
costinm: Take a look at the makefile, an alternative is to label the namespace as 'do not inject'.
We probably still want to have a 'delete' function that will cleanup the webhook - maybe part of istioctl
or operator.

bin/test.sh
```diff
@@ -65,13 +65,13 @@ if [ "$SKIP_SETUP" -ne 1 ]; then
     ISTIO_CONTROL=${ISTIO_CONTROL:-istio-control}
 
     kubectl -n bookinfo delete -f samples/bookinfo/platform/kube/bookinfo.yaml --ignore-not-found
-    kubectl -n bookinfo delete -f samples/bookinfo/networking/destination-rule-all-mtls.yaml --ignore-not-found
+    kubectl -n bookinfo delete -f samples/bookinfo/networking/destination-rule-all.yaml --ignore-not-found
```
costinm: In the makefile I am adding both DR and Policy to the simple test - I think we should recommend as best practice for all users to be explicit and include the desired settings for each namespace instead of assuming or relying on install options. 

It may be better to run bookinfo with mtls on - we do test mtls off with simple.

bin/install.sh
c0d1ngm0nk3y: We could simply remove the label, right? But the webhook is outdated anyway once the injector was deleted. So the operator sounds like a good place, imho it should be automatically removed.

bin/test.sh
c0d1ngm0nk3y: We just saw that the pipeline was broken after #42 - so we did the simplest fix after finding the problem. Currently, the pipeline "only" installs the default given by the [global|value].yaml.

c0d1ngm0nk3y:
> Any chance the feedback from concourse can be exposed in the status ?

That was the initial goal. @holgero , right? Currently, the pipeline runs with each new commit on master, meaning AFTER the merge when it is "too late".

holgero:
@c0d1ngm0nk3y yes, that is the idea. We will need to explore a bit how to configure the concourse instance that we have. E.g. its build logs should become visible to others and PR authors should be enabled to re-trigger the job. Also we might need to restrict who can trigger a run (can't simply run foreign code from every pull request without some sort of check. We probably need something like that ok_to_test mechanism).

PR #51, opened 2019-04-23
cleanup.sh: remove istio-env label from all namespaces

c0d1ngm0nk3y:
This makes the concourse pipeline more predictable as `install.sh` decides which namespace to use for `istio-control` based on `istio-env` labels of other namespaces (first in list).

### DIFF
--- a/bin/cleanup.sh
+++ b/bin/cleanup.sh
@@ -4,3 +4,8 @@
 for namespace in istio-system istio-control istio-control-master istio-ingress istio-telemetry; do
     kubectl delete namespace $namespace --wait --ignore-not-found
 done
+
+ACTIVE_NAMESPACES=$(kubectl get namespaces --no-headers -l istio-env -o=custom-columns=NAME:.metadata.name)
+for ns in $ACTIVE_NAMESPACES; do
+    kubectl label namespaces ${ns} istio-env-
+done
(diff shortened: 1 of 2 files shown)


bin/cleanup.sh
```diff
@@ -4,3 +4,8 @@
 for namespace in istio-system istio-control istio-control-master istio-ingress istio-telemetry; do
```
costinm: Can we label all iop-installed namespaces with a istio:something label ? I would expect at some point we'll run the control plane and other components in namespaces derived from the PR ( to parallelize )

bin/cleanup.sh
```diff
@@ -4,3 +4,8 @@
...
     kubectl delete namespace $namespace --wait --ignore-not-found
 done
+
```
costinm: Also maybe move the removal of the webhooks ?

c0d1ngm0nk3y: You mean not to use an `ownerReference` but to delete it in `cleanup.sh`?

But isn't the `ownerReference` better? You don't have the risk of having a webhook that is not working because the sidecar injector was removed.

To make this clear: when we use the `ownerReference` and delete the namespace here in `cleanup.sh` the webhook will be removed automatically.

bin/cleanup.sh
c0d1ngm0nk3y: We see your point, but don't think this should be part of this PR. We think this would require other changes as well, right?

bin/cleanup.sh
costinm: Interesting, didn't know this. Sounds cool. Can we do it for Galley validator as well (in istio/istio since that's generated)?

PR #60, opened 2019-04-23
Add prefix to values files

howardjohn:
Going to batch this to avoid 1000 of changes

fwiw I used this script in each dir to replace:
```
fixprefix() {
    prefix=${1:?prefix}
    find ./ -type f -exec sed -i -e "s/.Values/.Values.$prefix/g" {} \;
    find ./ -type f -exec sed -i -e "s/.Values.$prefix.global/.Values.global/g" {} \;
}
```


howardjohn:
Part of https://github.com/istio-ecosystem/istio-installer/issues/36

gateways/istio-egress/templates/_helpers.tpl
```diff
@@ -3,7 +3,8 @@
 Expand the name of the chart.
 */}}
 {{- define "gateway.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{ $gateway := index .Values "gateways" "istio-egressgateway" }}
```
costinm: Please get rid of this - no need to use a tempalate to get the name, we no longer have gateways.

gateways/istio-egress/templates/autoscale.yaml
```diff
@@ -1,4 +1,5 @@
-{{- if .Values.autoscaleMin }}
+{{ $gateway := index .Values "gateways" "istio-egressgateway" }}
```
howardjohn: Go template cannot have hyphen so we need this hack

costinm: We're no longer using this - each egress is 'a-la-carte', installed independently

howardjohn: I though the whole point of the prefix was for backwards compatibility with 1.1, which is configured via `gateways.istio-egressgateway`

gateways/istio-egress/values.yaml
```diff
@@ -1,87 +1,59 @@
...
-  sds:
-    enabled: false
-  proxy:
-    enableCoreDump: false
-    image: proxyv2
-
-  tracer:
-    zipkin:
-      address: zipkin.istio-system:9411
+gateways:
```
costinm: See above - each gateway goes to a different namespace, managed by a different team (possibly), with separate RBAC rules and secrets. 
That's the point of the a-la-carte and modular installer.

istio-control/istio-autoinject/templates/_helpers.tpl
```diff
@@ -3,7 +3,7 @@
 Expand the name of the chart.
 */}}
 {{- define "sidecar-injector.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- default .Chart.Name .Values.sidecarInjectorWebhook.nameOverride | trunc 63 | trimSuffix "-" -}}
```
costinm: This should also go - didn't clean the helpers, but should not be used. Helm leftover

gateways/istio-egress/templates/autoscale.yaml
costinm: Where would the hyphen be ? We use same thing for ingress and seems to work. 

We can make some changes - it is a new installer after all. It is a bit different from 1.0 to 1.1 in-place upgrade - the move from 1.1 to 1.2 should not have the same risks of accidentally breaking stuff, and since we move to different namespaces it is quite fine to change some names.

costinm: Backward compat where it makes sense and doesn't conflict with the new design. 

For security and operation reasons we want to isolate the components in different namespaces - so secrets, etc can be managed by different teams and deployed/operated independently.

costinm:
Well, not sure why my comment is in triplicate :-)

gateways/istio-egress/templates/autoscale.yaml
howardjohn: they are still separated the only difference is how they are configured through helm
--set gateways.egress.something vs --set egress.something.  the 2nd is probably better ux but 1st isn't so bad and is compatible.

howardjohn: Before we would set something like `.Values.something`, which a user would assign with `---set gateways.istio-egressgateway.something=whatever`. Now we need to address this at `.Values.gateways.istio-egressgateway.something` in the templates, which has the hyphen that breaks it. 

So we don't use the same for egress, as it doesn't have a prefix for values yet, which breaks everyones values.yaml files

istio-control/istio-autoinject/templates/_helpers.tpl
howardjohn: https://github.com/istio-ecosystem/istio-installer/issues/66

Will clean them in a follow up pr, too hard to track so many changes in one

gateways/istio-egress/templates/autoscale.yaml
howardjohn: Discussed offline, we will see if we can have both options work for now

PR #68, opened 2019-04-25
Add another a-la-carte test, and associated fixes for no-auth

costinm:
- injector template changed so it can be used with istioctl kubeinject file directly ( needs to be yaml starting with template:)
- added the associated tests and flags to allow running without certs

Quite a few remaining TODOs, but it appears the basics work.


test/noauth.mk
```diff
@@ -0,0 +1,59 @@
...
+# Will install 2 control planes, one with only discovery + ingress ( micro ), one with galley, discovery, telemetry
+run-test-noauth-micro: install-crds
+	bin/iop ${ISTIO_NS}-micro istio-discovery ${BASE}/istio-control/istio-discovery ${IOP_OPTS} \
+		--set global.controlPlaneSecurityEnabled=false --set useMCP=false --set plugins="health"
+	kubectl wait deployments istio-pilot -n ${ISTIO_NS}-micro --for=condition=available --timeout=${WAIT_TIMEOUT}
+	bin/iop ${ISTIO_NS}-micro istio-ingress ${BASE}/gateways/istio-ingress --set global.istioNamespace=${ISTIO_NS}-micro \
+	 	${IOP_OPTS} --set global.controlPlaneSecurityEnabled=false
+	kubectl wait deployments ingressgateway -n ${ISTIO_NS}-micro --for=condition=available --timeout=${WAIT_TIMEOUT}
+
+# TODO: pass meshConfigFile, injectConfigFile, valuesFile to test, or skip the kube-inject and do it manually (better)
```
howardjohn: Why can't it read the configmap, does it expect istio-system?

costinm: There is no configmap and no inject template - since we don't install the injector. 
We don't install the injector because there are no certs ( and we want to verify a-la-carte - injector is in no way required ).

howardjohn:
@costinm looks like the galley admission webhook isn't installed

PR #70, opened 2019-04-25
fix: update mixer config to match istio/istio

douglas-reid:
There were a large number of diffs around mixer config in the `mixer-telemetry` and `istio-policy` directories. This PR attempts to correct those issues.

In particular, this PR:
- updates the attribute manifests to match 1.1.x
- removes telemetry-related config from policy
- updates CRs to use the new-style config (only instances, handlers, rules)
- adds adapter config guards based on install options.

Helps to address #67.

### DIFF
--- a/istio-policy/templates/config.yaml
+++ b/istio-policy/templates/config.yaml
@@ -24,6 +24,10 @@ spec:
       valueType: STRING
     request.path:
       valueType: STRING
+    request.url_path:
+      valueType: STRING
+    request.query_params:
+      valueType: STRING_MAP
     request.reason:
       valueType: STRING
     request.referer:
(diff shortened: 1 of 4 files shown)


howardjohn:
In general looks good, not familiar with any of the mixer stuff though. Is this from release-1.1 or master branch?

douglas-reid:
@howardjohn i think i actually grabbed the source from master, but it should also be fully-aligned with 1.1 at this stage.

costinm:
I am working to add a test for 1.1 and upgrade, we'll fix if anything is broken.

douglas-reid:
OK. I can set these in a follow-on. I expect several as I work through testing with Mixer in any case.

PR #71, opened 2019-04-25
feat: add support for prometheus operator config

douglas-reid:
This PR is an attempt to add support for generating Prometheus Operator config to Istio. It is a port of https://github.com/istio/istio/pull/13407.

This PR assumes that the Prometheus Operator itself has been previously installed (example: `kubectl -n prom-operator apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/bundle.yaml`).

Motivations (Code Mauve):
- https://github.com/istio/istio/issues/10134
- https://github.com/istio/istio/issues/7528
- https://github.com/istio/istio/issues/5673
- https://github.com/istio/istio/issues/12915
- https://discuss.istio.io/t/istio-prometheus-to-scrape-metrics-from-service-specific-metrics-endpoint/926
- https://discuss.istio.io/t/how-to-update-prometheus-config/935
- https://discuss.istio.io/t/istio-1-1-daily-byo-prometheus-via-operator/418
- https://discuss.istio.io/t/is-envoy-metrics-not-pushed-via-statsd-after-1-1/1883

NOTE: I have not tested this fully yet. I have only copied it over from the other PR.

### DIFF
--- a/istio-telemetry/prometheus-operator/values.yaml
+++ b/istio-telemetry/prometheus-operator/values.yaml
@@ -1,18 +1,19 @@
-# Controls the default scrape interval used in the ServiceMonitors
-scrapeInterval: 15s
+prometheus:
+  # Controls the default scrape interval used in the ServiceMonitors
+  scrapeInterval: 15s
 
-# Enabling this option will generate a Prometheus resource, causing the operator
-# to create a prometheus deployment according to the generated spec. It will also
-# create a service definition, destination rules, cluster roles and bindings, and
-# a service account. Only use this option if you have not already created and configured
-# a prometheus resource and/or you desire a distinct prometheus resource for Istio.
-createPrometheusResource: false
-hub: docker.io/prom
-tag: v2.8.0
-retention: 6h
+  # Enabling this option will generate a Prometheus resource, causing the operator
+  # to create a prometheus deployment according to the generated spec. It will also
+  # create a service definition, destination rules, cluster roles and bindings, and
+  # a service account. Only use this option if you have not already created and configured
+  # a prometheus resource and/or you desire a distinct prometheus resource for Istio.
+  createPrometheusResource: false
+  hub: docker.io/prom
+  tag: v2.8.0
+  retention: 6h
 
-service:
-  annotations: {}
-  nodePort:
-    enabled: false
-    port: 32090
+  service:
+    annotations: {}
+    nodePort:
+      enabled: false
+      port: 32090
--- a/istio-telemetry/prometheus-operator/values.yaml
+++ b/istio-telemetry/prometheus-operator/values.yaml
@@ -1,0 +1,18 @@
+# Controls the default scrape interval used in the ServiceMonitors
+scrapeInterval: 15s
+
+# Enabling this option will generate a Prometheus resource, causing the operator
+# to create a prometheus deployment according to the generated spec. It will also
+# create a service definition, destination rules, cluster roles and bindings, and
+# a service account. Only use this option if you have not already created and configured
+# a prometheus resource and/or you desire a distinct prometheus resource for Istio.
+createPrometheusResource: false
+hub: docker.io/prom
+tag: v2.8.0
+retention: 6h
+
+service:
+  annotations: {}
+  nodePort:
+    enabled: false
+    port: 32090
--- a/istio-telemetry/prometheus-operator/Chart.yaml
+++ b/istio-telemetry/prometheus-operator/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+description: A Helm chart for Kubernetes
+name: prometheus-operator
+version: 1.1.0
+appVersion: 2.3.1
+tillerVersion: ">=2.7.2"
--- a/istio-telemetry/prometheus-operator/templates/_affinity.tpl
+++ b/istio-telemetry/prometheus-operator/templates/_affinity.tpl
@@ -1,0 +1,86 @@
+{{/* affinity - https://kubernetes.io/docs/concepts/configuration/assign-pod-node/ */}}
+
+{{- define "nodeaffinity" }}
+  nodeAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+    {{- include "nodeAffinityRequiredDuringScheduling" . }}
+    preferredDuringSchedulingIgnoredDuringExecution:
+    {{- include "nodeAffinityPreferredDuringScheduling" . }}
+{{- end }}
+
+{{- define "nodeAffinityRequiredDuringScheduling" }}
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: beta.kubernetes.io/arch
+          operator: In
+          values:
+        {{- range $key, $val := .Values.global.arch }}
+          {{- if gt ($val | int) 0 }}
+          - {{ $key }}
+          {{- end }}
+        {{- end }}
+{{- end }}
+
+{{- define "nodeAffinityPreferredDuringScheduling" }}
+  {{- range $key, $val := .Values.global.arch }}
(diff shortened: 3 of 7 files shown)


istio-telemetry/prometheus-operator/values.yaml
```diff
@@ -0,0 +1,18 @@
+# Controls the default scrape interval used in the ServiceMonitors
```
costinm: Please add a top-level "prometheus:", for consistency and compat with 1.1.
Can you do a diff with the values from 1.1, and maybe add a doc to any setting that is new, and 
a commented out setting if it was deleted ? 
( we need to do this for all components )

douglas-reid:
Note to self: need to update `istio-component-monitor` service monitor config, as some add/or all components may no longer be using a `http-monitoring` named port.

douglas-reid:
PTAL.  I've moved the `values.yaml` options under a top-level `prometheus` tag, as requested.

I added some very high-level "tests" to the Makefile to validate the right config is generated, etc. I'm not sure if that is what is being sought, however.

douglas-reid:
FWIW, as far as I can tell, none of the test failures are related to these changes.

douglas-reid:
Friendly ping on review.

Makefile
```diff
@@ -235,6 +235,19 @@ install-policy:
...
+install-prometheus-operator:
+	# installs the prom operator in the default namespace
+	kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/bundle.yaml
+	kubectl wait --for=condition=available --timeout=${WAIT_TIMEOUT} deploy/prometheus-operator
+
+# This target expects that the prometheus operator (and its CRDs have already been installed).
+# It is provided as a way to install Istio prometheus operator config in isolation.
+install-prometheus-operator-config:
+	kubectl create ns ${ISTIO_NS} || /bin/true
+	# NOTE: we don't use iop to install, as it defaults to `--prune`, which is incompatible with the prom operator (it prunes the stateful set)
```
howardjohn: How come? Does the statefulset not get the labels?

Makefile
```diff
@@ -235,6 +235,19 @@ install-policy:
...
+# This target should only be used in situations in which the prom operator has not already been installed in a cluster.
+install-prometheus-operator:
+	# installs the prom operator in the default namespace
+	kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/bundle.yaml
+	kubectl wait --for=condition=available --timeout=${WAIT_TIMEOUT} deploy/prometheus-operator
+
+# This target expects that the prometheus operator (and its CRDs have already been installed).
+# It is provided as a way to install Istio prometheus operator config in isolation.
+install-prometheus-operator-config:
+	kubectl create ns ${ISTIO_NS} || /bin/true
```
howardjohn: can you change to `true` instead of `/bin/true`? Needed for mac compatibility apparently

Makefile
douglas-reid: yeah, the operator doesn't apply the labels to the stateful set, it turns out.

douglas-reid:
Awesome!  Thanks for the reviews and help..

davidham:
Hi! Sorry to comment on a closed PR, but is there any documentation on how to use this? I am adding Istio to a cluster that already has prometheus-operator installed, and I'd love to have these metrics and dashboards be able to be sent to my existing install. Thank you!

howardjohn:
@davidham the charts are at https://github.com/istio/installer/tree/master/istio-telemetry/prometheus-operator

So you can just `helm template`/`helm install` those and it should work. If not, please let us know

davidham:
Ooooooooh nice! So if I installed Istio per the docs, should I just update my install to be `prometheus.enabled = false` and then install the prom-operator you mentioned above? Can I still make use of the grafana dashboards that come with Istio?

howardjohn:
Yep, you might need to configure grafana to point to the right prometheus though since its probably not on the same address as grafana expects out of the box

Martinlcksybo:
What is the name of the repository and chart? What's the helm command I should run?

davidham:
@howardjohn So: I wanted to install the Istio grafana dashboards, but use them in my existing prometheus/grafana setup from a separate install of `stable/prometheus-operator`. I installed the prom-operator from istio/installer/telemetry, and it seems to have worked because it installed ServiceMonitors for all the Istio stuff, and heir metrics show up in Prometheus. So far so good!

I also installed Grafana from istio/installer/telemetry, and this install worked also (though I had to get `fix_datasources.sh` to work first). It installed all the dashboard configmaps, but they aren't being picked up by my Grafana. Post-install I added the `grafana_dashboard` label to each one, and I updated the istio-grafana configmap to point to my main prometheus service. 

I looked at the logs for the grafana-sc-dashboard container, and i can see it reading all the configmaps, but it doesn't seem to be reading the ones in istio-system. Is there anything else I have to do to get these dashboards to work?

davidham:
☝️ Got this to work, I had neglected to create a rolebinding for Prometheus in the `istio-system` namespace. Ince I did this it worked fine.

mabushey:
> So you can just `helm template`/`helm install` those and it should work. If not, please let us know

Is there a way to just get the Yaml? Helm is too complicated for me. I can easily use SED or VIM to modify a yaml file. I don't understand layering on massive levels of complexity on top of something so simple and elegant as yaml config.

douglas-reid:
@mabushey i don't know that we have copy of these yamls with all the default values populated, but generating it yourself should be straightforward, even without helm. 

you could go the location of the template files (already linked above), grab the yaml that you want, copy it to wherever, and sed/vim away. you'll need to substitute out all the template stuff (contained within '{{' and '}}') with the values you want, of course.

fwiw, bending helm to your will can definitely be frustrating, but, for this specific generation task, just running `helm template` might be the simplest approach in the short term.

mabushey:
@douglas-reid 

I only use `helm template`, but usually it complains that the chart is not gziped or it throws an error like `Error: render error in "prometheus/templates/tests/test-prometheus-connection.yaml": template: prometheus/templates/tests/test-prometheus-connection.yaml:1:14: executing "prometheus/templates/tests/test-prometheus-connection.yaml" at <.Values.global.enableHelmTest>: nil pointer evaluating interface {}.enableHelmTest`

douglas-reid:
that happens when you set the chart to the lower level chart. hooray, helm!

the "secret" is to use the top-level chart and then (ugh) `-x <template>` for each template in the underlying chart (which here might just be one, if not, then both).

haf-afa:
@douglas-reid 

> i don't know that we have copy of these yamls with all the default values populated

This would be very useful. It would allow us to use kustomize.

> you could go the location of the template files (already linked above), grab the yaml that you want, copy it to wherever, and sed/vim away. you'll need to substitute out all the template stuff (contained within {{ and }}) with the values you want, of course.

This what that looks like for prometheus-operator: https://gist.github.com/haf-afa/575ed9f546037ea99c5b9e97f75b9c83#file-kustomization-yaml It took 2 days to make it work just for the operator with a stand-alone configured Grafana. Istio does both Grafana and Prom-Op, so I expect another two days for that. I've spent half a day so far, and gotten the dashboards to show up with the kiwigrid-sidecar fetching of configmaps. But it's not enough; I also have to modify the Prometheus configuration to match that of this repo, so that the values are populated in the Prometheus TSDB instance.

@mabushey I'd be happy to collaborate with you on getting this stack set up with plain Kubernetes files, the Prom Operator, Grafana and Istio. It would seem we have overlapping use-cases. I tried urging those good folk to do the same here https://github.com/coreos/prometheus-operator/issues/2853 - see the screenshots I posted; there were 163 different variables in that helm chart that could be tweaked when I counted; not counting the CRD definitions that could be inserted.

Issue #54, opened 2019-04-23
Enable CNI


sdake:
Please use this Epic to track any issues encountered with the CNI.

tiswanso:
FYI:  [istio CNI install instructions](https://istio.io/docs/setup/kubernetes/additional-setup/cni/)

On initial review of integrating CNI here:
- [ ] add a version of istio-cni helm chart to this repo
- [ ] modify `bin/install.sh`
    - add a command flag for `install_cni`
    - what should `install-all` do w.r.t. `install_cni`?
       - proposal:  have `install_cni` be a separate step and modify `install_control` to detect whether `istio-cni` is installed to set helm value for `istio_cni.enabled=true` when sidecar injector configmap is created (`iop istio-autoinject`)
    - how to detect which k8s env we're on?  e.g. `istio-cni` on GKE requires a specific setting for `cni-bin-dir`

PR #74, opened 2019-04-25
Add Istio cni support

tiswanso:
- Added istio/cni helm charts.  Made each k8s object its own template.
- scoped all istio-cni helm vars to `cni` top level key
- Replaced `istio_cni.enable` helm var with `global.proxy.injectionControl` (values = `cni`, `init`)
- updated README with istio-cni install info

Fixes: #54

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,9 +238,9 @@ the default disabled, test it, and move the default from istio-system to istio-c
 
 
 ```bash
-    # determine whether the istio-cni is installed    
+    # ENABLE_CNI is set to true if istio-cni is installed    
     iop istio-control istio-autoinject $IBASE/istio-control/istio-autoinject --set enableNamespacesByDefault=true \
-        --set global.proxy.interceptionControl=CNI
+        --set istio_cni.enabled=${ENABLE_CNI}
     
     # Second auto-inject using master version of istio
     # Notice the different options
--- a/README.md
+++ b/README.md
@@ -164,10 +164,22 @@ Access to the security namespace and istio-system should be highly restricted.
 
 ## Install Istio-CNI
 
-TODO. This is an optional step - CNI must run in a dedicated namespace, it is a 'singleton' and extremely 
+This is an optional step - CNI must run in a dedicated namespace, it is a 'singleton' and extremely 
 security sensitive. Access to the CNI namespace must be highly restricted. 
 
-It is possible to add Istio-CNI later, and gradually migrate. 
+**NOTE:** The environment variable `ISTIO_CLUSTER_ISGKE` is assumed to be set to `true` if the cluster
+is a GKE cluster.
+
+```bash
+ISTIO_CNI_ARGS=
+# TODO: What k8s data can we use for this check for whether GKE?
+if [[ "${ISTIO_CLUSTER_ISGKE}" == "true" ]]; then
+    ISTIO_CNI_ARGS="--set cniBinDir=/home/kubernetes/bin"
+fi
+bin/iop istio-cni istio-cni $IBASE/istio-cni/ ${ISTIO_CNI_ARGS}
+```
+
+TODO. It is possible to add Istio-CNI later, and gradually migrate. 
 
 ## Install Control plane 
 
@@ -226,7 +238,11 @@ the default disabled, test it, and move the default from istio-system to istio-c
 
 
 ```bash
-    iop istio-control istio-autoinject $IBASE/istio-control/istio-autoinject --set enableNamespacesByDefault=true
+    # determine whether the istio-cni is installed
+    cni_pods=$(kubectl get pods -n istio-cni -l k8s-app=istio-cni-node -o "jsonpath={.items..metadata.name}")
+    
+    iop istio-control istio-autoinject $IBASE/istio-control/istio-autoinject --set enableNamespacesByDefault=true \
+        ${cni_pods:+ --set istio_cni.enabled=true}
     
     # Second auto-inject using master version of istio
     # Notice the different options
--- a/README.md
+++ b/README.md
@@ -238,11 +238,9 @@ the default disabled, test it, and move the default from istio-system to istio-c
 
 
 ```bash
-    # determine whether the istio-cni is installed
-    cni_pods=$(kubectl get pods -n istio-cni -l k8s-app=istio-cni-node -o "jsonpath={.items..metadata.name}")
-    
+    # determine whether the istio-cni is installed    
     iop istio-control istio-autoinject $IBASE/istio-control/istio-autoinject --set enableNamespacesByDefault=true \
-        ${cni_pods:+ --set istio_cni.enabled=true}
+        --set global.proxy.interceptionControl=CNI
     
     # Second auto-inject using master version of istio
     # Notice the different options
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -126,7 +126,7 @@ function install_cni() {
     if [[ "${ISTIO_CLUSTER_ISGKE}" == "true" ]]; then
         ISTIO_CNI_ARGS="--set cniBinDir=/home/kubernetes/bin"
     fi
-    bin/iop istio-cni istio-cni $IBASE/istio-cni/ -c $IBASE/istio-cni/values.yaml ${ISTIO_CNI_ARGS} 
+    bin/iop istio-cni istio-cni $IBASE/istio-cni/ -f $IBASE/istio-cni/values.yaml ${ISTIO_CNI_ARGS}
     kubectl rollout status ds istio-cni-node -n istio-cni --timeout=$WAIT_TIMEOUT
 }
 
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -23,21 +23,6 @@ function step() {
     echo "${METHOD}ing $1"
 }
 
-# utility to check for pod existence
-function pod_exists() {
-    local ns=$1
-    shift
-    local labelmatch=$1
-    shift
-
-    found_pods=$(kubectl get pods -n ${ns} ${labelmatch:+ -l ${labelmatch}} -o "jsonpath={.items..metadata.name}")
-    if [[ "$found_pods" != "" ]]; then
-        true
-    else
-        false
-    fi
-}
-
 # Install istio crds
 
 function install_crds() {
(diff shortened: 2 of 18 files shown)


tiswanso:
NOTE: the helm chart `istio-cni` is a straight copy from istio/cni

I'm not sure if the kustomize step strips out the serviceaccount stuff.  Not sure if it can handle multiple part yaml template like istio-cni has... looks like the rest of the templates have separate files for each k8s object.  Probably the istio-cni chart is an antipattern.  Will try moving stuff into files.

README.md
```diff
@@ -157,10 +157,22 @@ Access to the security namespace and istio-system should be highly restricted.
...
-TODO. This is an optional step - CNI must run in a dedicated namespace, it is a 'singleton' and extremely 
+This is an optional step - CNI must run in a dedicated namespace, it is a 'singleton' and extremely 
 security sensitive. Access to the CNI namespace must be highly restricted. 
 
-It is possible to add Istio-CNI later, and gradually migrate. 
+**NOTE:** The environment variable `ISTIO_CLUSTER_ISGKE` is assumed to be set to `true` if the cluster
+is a GKE cluster.
+
+```bash
+ISTIO_CNI_ARGS=
```
costinm: Can we determine this from kubectl cluster-info ? If yes, we can add it to the iop script ( and the future go binary that will do the same thing ). 

Also - wondering if there is any detection that can be done in cni itself...

(for future PR)

README.md
```diff
@@ -219,7 +231,11 @@ the default disabled, test it, and move the default from istio-system to istio-c
 
 
 ```bash
-    iop istio-control istio-autoinject $IBASE/istio-control/istio-autoinject --set enableNamespacesByDefault=true
+    # determine whether the istio-cni is installed
```
costinm: I would avoid this - the recommendation is for user to have a user-values.yaml where he specifies various settings.

For upgrade and gradual adoption we want CNI to work nicely with injected pods with iptables.

README.md
```diff
@@ -219,7 +231,11 @@ the default disabled, test it, and move the default from istio-system to istio-c
...
+    cni_pods=$(kubectl get pods -n istio-cni -l k8s-app=istio-cni-node -o "jsonpath={.items..metadata.name}")
+    
+    iop istio-control istio-autoinject $IBASE/istio-control/istio-autoinject --set enableNamespacesByDefault=true \
```
costinm: We should have a new option 'disable iptables' or similar - since NONE interception mode is also iptable-less ( doesn't need the extra container ).

bin/install.sh
```diff
@@ -17,11 +17,27 @@
 cd  "$( dirname "${BASH_SOURCE[0]}" )/.."
 export IBASE="$(pwd)"
 export WAIT_TIMEOUT=${WAIT_TIMEOUT:-5m}
+ISTIO_CLUSTER_ISGKE=${ISTIO_CLUSTER_ISGKE:-false}
 
 function step() {
     echo "${METHOD}ing $1"
 }
 
+# utility to check for pod existence
```
costinm: Please remove this part - should be based on values only. It may mess up with other things - we generate a yaml and apply kustomize and may apply it to a different cluster.

bin/iop
```diff
@@ -84,8 +89,9 @@ function render_kube_yaml() {
 }
 
 if [ "$1" == "" ] ; then
-    echo "iop NAMESPACE COMPONENT CONFIGDIR [-t|-d] [HELM_TEMPLATE_OPTIONS]"
+    echo "iop NAMESPACE COMPONENT CONFIGDIR [-c <cfgfile.yaml>] [-t|-d] [HELM_TEMPLATE_OPTIONS]"
```
costinm: Why ? You can pass additional -f to override the global.yaml and values.yaml.
Not sure it's worth the complexity, any reason -f myglobal.yaml won't work ?

istio-cni/values.yaml
```diff
@@ -0,0 +1,20 @@
+hub: gcr.io/istio-release
```
costinm: There is a bug on all components we are trying to fix - as we refactored to top-level templates we broke backward compat with 1.1 values.yaml, where settings were prefixed by the component name.

Should be:
cni:
   hub: ...

and in templates {{ .Values.cni.hub }}

README.md
tiswanso: I'll see what I can figure out with cluster-info.  The problem with trying to do it in the CNI itself is that the `istio-cni-node` pod needs to volume mount the host's cni-bin-dir so it can copy in the Istio CNI binary.  So, to discover the cni-bin-dir or which cloud-platform we're on we'd need to have something like a k8s operator for CNI.

README.md
tiswanso: Are you meaning we should rename `istio_cni.enabled` to something that more represents "don't include the iptables init-container"?

bin/iop
tiswanso: Ok yeah... didn't recall I could pass an additional `-f ...`.  Will remove

README.md
costinm: Can we mount both ? With optional=true maybe ? Anything we can do to simplify the UX.
But we can do it in separate PRs - it's fine to use a values.yaml setting for now - doesn't need to be 
part of the iop script.

README.md
costinm: Yes, absolutely. We already have this case for NONE (whitebox). 
We can leave it for separate PR - but please remove the cni_pods thing - the config generation shouldn't depend on access to k8s. If the values have it set - we use it.

What I think is critical is upgrade and gradual adoption story - we must make sure that CNI 
will work fine if some pods have injection enabled ( or use manual kube-inject ).

README.md
tiswanso: AFAIK, there's no "optional" setting for host mounts.  We should be able to use "DirectoryOrCreate" and check in the install script if the directory is empty or not before installing the binary to detect.  Will create a CNI PR for that shortly.

README.md
tiswanso: Do you think I should explain the enable of CNI more in words rather than a jsonpath example?

It seems like the point of the README is to explicitly show some basic `iop` usage for the specific features being installed.  So, I was just showing how they could detect the istio-cni and set flag to enabled/disable the init container.

README.md
tiswanso: Auto or manual inject doesn't matter as long as the sidecar-injector-configmap is setup correctly (with initContainer disabled).  For the case where the user wants some pods to get their iptables setup by the CNI and some to be setup by the initContainer, then multiple configmaps need to be created and manual injection should point at the appropriate configmap.  I'm not sure whether that exact scenario is clearly explained in docs.

bin/install.sh
tiswanso: Ok... I was thinking that this is going to be the basis for the operator and this is what the operator should do, IMO.

bin/iop
tiswanso: removing `-c ...`

costinm:
Ping ?

costinm:
Can you rebase ? Seems some tests were fixed.

README.md
costinm: Got it, thanks for confirming.

For a separate PR maybe we can add some detection (kubect cluster info or something)

README.md
costinm: So if I have an existing pod, manually injected, and I add CNI - what will happen ? Will both iptables script be run ? 

Is there a documented way (annotation, etc) to disable CNI iptables ?

global.yaml
```diff
@@ -72,6 +72,12 @@ global:
         cpu: 2000m
         memory: 128Mi
 
+    # proxy.interceptionControl
+    # values:
+    #  init -- the istio-init container will setup the proxy traffic interception
+    #  cni -- the CNI will setup the proxy traffic interception (use this with istio-cni)
+    interceptionControl: init
```
costinm: @linsun what do you think of this API ? 

I like that we can extend it later.

tiswanso: I'm having a problem with the sidecar inject go template trying to check if this option is set and equal to init OR if the old option istio_cni.enabled is set to false.  So, I think I should revert this new option from this PR and keep the old one and propose something new in a follow-on.

README.md
tiswanso: NOTE:  CNI will only operate on newly created pods.  

If you have an injected manifest that has the istio-init container in it AND you have the istio-cni installed, the worst case is that pod netns iptables rules will be added by istio-cni prior to the istio-init container running and when the istio-init container runs it will clobber and re-apply all the rules.  So, unless the istio-cni and istio-init iptables setup scripts get way out of sync the user would not notice any change in resulting istio-proxy behavior.  

The istio-cni pays attention to the same annotations as the sidecar injector so, `sidecar.istio.io/inject` set to false would disable the istio-cni from doing its thing BUT it would also disable the sidecar injector.  You could toggle the value after manual inject on a pod with istio-init to avoid the istio-cni doing anything, though.  But that seems awkward.  

The istio-cni can be disabled on a namespace level in its CNI level config so that's another option unrelated to sidecar injection.

costinm:
Also - since interceptionMode can be specified per pod via annotation - will allow gradual adoption of CNI.

I think it is required that the CNI agent skips iptables init in some cases, we want all features of istio to be 'a-la-carte' and support gradual adoption ( few test apps first, then roll out to more, etc) and give users some control. 

Another factor to keep in mind is that with the multi-namespace support there is no longer a 'global' config - you can have istio-prod set with a certain values, and istio-testing use a different one. 
CNI and citadel agent are globals - so they'll run on all nodes - but they still need to respect the settings
in various environments.

tiswanso:
@costinm -- IMO, we should have as follow-on PRs/wiki/doc updates for the larger discussion topics.  Agree?  

I think the disconnect is that the API knob we're discussing is actually just for whether the injection template pushes in a istio-init initContainer or not.  It does not control whether the istio-cni is installed or any configuration for the istio-cni. 

The redirection type is separate as BOTH istio-init and istio-cni setup TPROXY or the fullstack redirect in iptables.  It just happens that someone used the NONE value as a knob for disabling istio-init also.

RE: params to CNI -- Changing the CNI to start looking at more than per-pod annotations would be a bigger enhancement to the CNI that hasn't been started or really evaluated.  As mentioned in the prior comment there are multiple ways for coexistence but you're right that needs some examples/documentation but I'd like to get that in follow-up.

README.md
costinm: That's great - you are right istio-init can clober the rules, so it can override. I think this is sufficient, we can improve later if anyone asks..

bin/install.sh
```diff
@@ -99,6 +100,17 @@ function install_telemetry() {
...
 }
 
+# Install Istio CNI
+function install_cni() {
+    step "cni.."
+    ISTIO_CNI_ARGS=
+    if [[ "${ISTIO_CLUSTER_ISGKE}" == "true" ]]; then
+        ISTIO_CNI_ARGS="--set cniBinDir=/home/kubernetes/bin"
+    fi
+    bin/iop istio-cni istio-cni $IBASE/istio-cni/ ${ISTIO_CNI_ARGS}
```
costinm: Should we place CNI in istio-system, since it has the same level of security impact as Citadel ? 
May also be useful if in future we merge the node agents. But can be done in separate PR.

PR #79, opened 2019-04-26
Fix configValidation setting

howardjohn:
It was previously at the global level (no effect) and turned off by
default (should be on by default).

@costinm this should be on by default right? If not we will probably get tons of issues from users appliyng invalid config

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -228,9 +228,6 @@ global:
   # the desired values.
   defaultNodeSelector: {}
 
-  # Whether to perform server-side validation of configuration.
-  configValidation: true
-
   # Custom DNS config for the pod to resolve names of services in other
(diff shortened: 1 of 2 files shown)


costinm:
Interesting question - initial thinking was to optimize for upgrade. I'm not sure upgrade works well if this is on, it may conflict. 

How about: leave it off with a bug filed to turn it on for new installs.
Once we have a test for upgrade from 1.1 to 1.2 - we can move it back  to on in all cases.
Or if you can test even manually upgrade works.

The problem is the global config, I think both galley in 1.1/istio-system and new one will fight each other.

costinm:
Fixed this as part of my previous PR ( tests were failing )

PR #80, opened 2019-04-27
support multiple app namespaces.

morvencao:
Context: support watching multiple namespaces for pilot

This is a followup PR for https://github.com/istio/istio/pull/13635

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -21,9 +21,6 @@ global:
   istioNamespace: istio-control
   configNamespace: istio-control
 
-  # Applications namespace list pilot manages, separated by comma, if it contains multiple namespaces.
-  appNamespace: ""
-
   # Telemetry namespace, including tracing.
   telemetryNamespace: istio-telemetry
 
--- a/global.yaml
+++ b/global.yaml
@@ -21,6 +21,9 @@ global:
   istioNamespace: istio-control
   configNamespace: istio-control
 
+  # Applications namespace list pilot manages, separated by comma, if it contains multiple namespaces.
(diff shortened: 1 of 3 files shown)


global.yaml
```diff
@@ -21,6 +21,9 @@ global:
   istioNamespace: istio-control
   configNamespace: istio-control
 
+  # Applications namespace list pilot manages, separated by comma, if it contains multiple namespaces.
+  appNamespace: ""
```
sdake: shouldn't this be a pilot specific config in the pilot specific values.yaml file?

costinm:
As in the other PR implementing this: 
- please not comma separate, use proper yaml ( i.e. a list )
- I'm not sure it's pilot only - mixer and all other components may need to follow the same list consistently. I suspect injector in particular. 

I'm not sure there is a use case for each component watching/using different namespaces.

costinm:
You can still use the env variable in pilot for now - until we move this to mesh config and pilot gets it from there, but as API it should be exposed as a list, not comma separate string.  

We need to be very careful with all new fields and changes in values.yaml - and treat it as a highly visible API.

global.yaml
costinm: +1, and also please no comma separate list in yaml !

morvencao:
@costinm Not sure if this is OK, but if we want to use the env variable, then we need to transform the list to string.

pinkertonpg:
The PR these changes were made for was never merged as far as I can tell: https://github.com/istio/istio/pull/13635

Should these changes be reverted?

PR #81, opened 2019-04-29
add probe for citadel.

morvencao:


security/citadel/templates/deployment.yaml
```diff
@@ -55,6 +55,12 @@ spec:
           {{- if .Values.security.trustDomain }}
             - --trust-domain={{ .Values.security.trustDomain }}
           {{- end }}
+          livenessProbe:
+            httpGet:
+              path: /version
+              port: {{ .Values.global.monitoringPort }}
```
sdake: This is not in globals.yaml.  I believe we decided to hardcode the monitoring port for 1.2.

sdake:
A grep of istio-installer shows 15014 port is hardcoded.

security/citadel/templates/deployment.yaml
costinm: Yes, some of the settings we never tested and likely don't support should be replaced by default values.

We support kustomize for cases user really needs a custom value.

morvencao:
@sdake Thanks, replace probe port with hard coded value.

PR #88, opened 2019-04-30
mount certs if config validation is enabled

costinm:


.mergify.yml
```diff
@@ -1,9 +1,13 @@
 pull_request_rules:
   - name: automatic merge on CI success and review
     conditions:
-      - status-success=continuous-integration/circleci/pr
+      - status-success=cla/google
+      - status-success=ci/circleci
```
howardjohn: I think it needs to exactly match the name? It is (hopefully) fixed this in master so if you rebase it may work

howardjohn:
@costinm can you rebase?

test/install.mk
```diff
@@ -21,11 +21,13 @@ run-build: dep
 	bin/iop ${ISTIO_NS} istio-telemetry ${BASE}/istio-telemetry/mixer-telemetry -t > ${OUT}/release/istio-telemetry.yaml
 	bin/iop ${ISTIO_NS} istio-telemetry ${BASE}/istio-telemetry/prometheus -t > ${OUT}/release/istio-prometheus.yaml
 	bin/iop ${ISTIO_NS} istio-telemetry ${BASE}/istio-telemetry/grafana -t > ${OUT}/release/istio-grafana.yaml
-	bin/iop ${ISTIO_NS} istio-policy ${BASE}/istio-policy -t > ${OUT}/release/istio-policy.yaml
+	#bin/iop ${ISTIO_NS} istio-policy ${BASE}/istio-policy -t > ${OUT}/release/istio-policy.yaml
 	#bin/iop ${ISTIO_NS} istio-cni ${BASE}/istio-cni -t > ${OUT}/release/istio-cni.yaml
 	# TODO: generate single config (merge all yaml)
 	# TODO: different common user-values combinations
 	# TODO: apply to a local kube apiserver to validate against k8s
+	# Short term: will be checked in - for testing apply -k
```
howardjohn: Should we be running this on each change and checking in changes?

.mergify.yml
costinm: rebased.

test/install.mk
costinm: yes, if it works. Will need some bot to do it - I'm not sure if checkin back to this git or push in some place.

costinm: But if it works - it will be really nice, download nothing is the simplest we can get.

.mergify.yml
```diff
@@ -12,4 +12,4 @@ pull_request_rules:
     actions:
       merge:
         method: squash
-        
+        strict: false
```
howardjohn: Seems like `strict: true` is actually a good idea, especial with installer since it is pretty easy for two changes to be incompatible and get merged in. Prow already does `strict: true` basically on istio/istio. Although maybe mergify's version of it isn't as good

PR #91, opened 2019-05-01
Sync galley templates with master

howardjohn:

### DIFF
--- a/istio-control/istio-config/templates/_podDisruptionBudget.tpl
+++ b/istio-control/istio-config/templates/_podDisruptionBudget.tpl
@@ -1,0 +1,8 @@
+{{- define "podDisruptionBudget.spec" }}
+{{- if .minAvailable }}
+  minAvailable: {{ .minAvailable }}
+{{- end }}
+{{- if .maxUnavailable }}
+  maxUnavailable: {{ .maxUnavailable }}
+{{- end }}
+{{- end }}
--- a/istio-control/istio-config/templates/deployment.yaml
+++ b/istio-control/istio-config/templates/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: istio-galley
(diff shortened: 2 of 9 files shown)


istio-control/istio-config/templates/_podDisruptionBudget.tpl
```diff
@@ -0,0 +1,8 @@
+{{- define "podDisruptionBudget.spec" }}
```
costinm: I'm  trying to minimize templating - you can just inline it, it's not used in any other place.

istio-control/istio-config/templates/deployment.yaml
```diff
@@ -20,20 +24,24 @@ spec:
...
+{{- if contains "/" .Values.galley.image }}
+          image: "{{ .Values.galley.image }}"
+{{- else }}
+          image: "{{ .Values.global.hub }}/{{ .Values.galley.image | default "galley" }}:{{ .Values.global.tag }}"
+{{- end }}
           imagePullPolicy: {{ .Values.global.imagePullPolicy | default "Always" }}
           ports:
           - containerPort: 443
-          - containerPort: {{ .Values.global.monitoringPort | default "15014" }}
+          - containerPort: {{ .Values.global.monitoringPort }}
```
costinm: Can we leave it ? monitoringPort is another setting we may deprecate, there is no reason to customize it or plan to test such customization.

istio-control/istio-config/templates/service.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   ports:
   - port: 443
     name: https-validation
-  - port: 15014
+  - port:  {{ .Values.global.monitoringPort }}
```
costinm: Same. If we can hardcode 9901 and 443, there is no reason to customize monitoringPort.

istio-control/istio-config/templates/deployment.yaml
```diff
@@ -20,20 +24,24 @@ spec:
...
+{{- else }}
+          image: "{{ .Values.global.hub }}/{{ .Values.galley.image | default "galley" }}:{{ .Values.global.tag }}"
+{{- end }}
           imagePullPolicy: {{ .Values.global.imagePullPolicy | default "Always" }}
           ports:
           - containerPort: 443
-          - containerPort: {{ .Values.global.monitoringPort | default "15014" }}
+          - containerPort: 15014
           - containerPort: 15019
+          - containerPort: 9901
```
costinm: This moved to the sidecar.

howardjohn: So we don't need this port now?

PR #92, opened 2019-05-01
Fix minikube install to use Make

howardjohn:
Minikube install doesn't get right environment variables because it is
calling the script directly, causing failures.


howardjohn:
@costinm still fails but during the actual tests

PR #93, opened 2019-05-01
Sync sidecar injector with master

howardjohn:

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -402,3 +402,8 @@ global:
   # This field is set to false by default, so 'helm template ...'
   # will ignore the helm test yaml files when generating the template
   enableHelmTest: false
+
+# If set, no iptable init will be added. It assumes CNI is installed.
+# TODO: rename to 'enableIptables' or add 'interceptionMode: CNI'
+istio_cni:
+  enabled: false
--- a/global.yaml
+++ b/global.yaml
@@ -92,9 +92,13 @@ global:
     # Configure the access log for sidecar to JSON or TEXT.
     accessLogEncoding: TEXT
 
-    # Log level for proxy, applies to gateways and sidecars.  If left empty, "warning" is used.
+    # Log level for proxy, applies to gateways and sidecars.
     # Expected values are: trace|debug|info|warning|error|critical|off
-    logLevel: ""
+    logLevel: warning
+
+    # Per Component log level for proxy, applies to gateways and sidecars. If a component level is
+    # not set, then the global "logLevel" will be used.
+    componentLogLevel: "misc:error"
 
     # Configure the DNS refresh rate for Envoy cluster of type STRICT_DNS
     # 5 seconds is the default refresh rate used by Envoy
@@ -403,7 +407,3 @@ global:
   # will ignore the helm test yaml files when generating the template
   enableHelmTest: false
 
-# If set, no iptable init will be added. It assumes CNI is installed.
-# TODO: rename to 'enableIptables' or add 'interceptionMode: CNI'
-istio_cni:
-  enabled: false
--- a/istio-control/istio-autoinject/files/injection-template.yaml
+++ b/istio-control/istio-autoinject/files/injection-template.yaml
@@ -11,7 +11,8 @@ template: |
   {{- else }}
     image: "{{ .Values.global.hub }}/{{ .Values.global.proxy_init.image }}:{{ .Values.global.tag }}"
   {{- end }}
-    args:
+    command:
+    - "/usr/local/bin/istio-iptables.sh"
     - "-p"
     - "{{ .MeshConfig.ProxyListenPort }}"
     - "-u"
@@ -87,9 +88,9 @@ template: |
     - --domain
     - $(POD_NAMESPACE).svc.{{ .Values.global.proxy.clusterDomain }}
     - --configPath
-    - "{{ .ProxyConfig.ConfigPath }}"
+    - "/etc/istio/proxy"
     - --binaryPath
-    - "{{ .ProxyConfig.BinaryPath }}"
+    - "/usr/local/bin/envoy"
     - --serviceCluster
     {{ if ne "" (index .ObjectMeta.Labels "app") -}}
     - "{{ index .ObjectMeta.Labels `app` }}.$(POD_NAMESPACE)"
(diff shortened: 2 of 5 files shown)


global.yaml
```diff
@@ -402,3 +402,8 @@ global:
   # This field is set to false by default, so 'helm template ...'
   # will ignore the helm test yaml files when generating the template
   enableHelmTest: false
+
+# If set, no iptable init will be added. It assumes CNI is installed.
+# TODO: rename to 'enableIptables' or add 'interceptionMode: CNI'
+istio_cni:
```
costinm: This should be under values.yaml - global should only have the global settings.

howardjohn: This is how it works in Istio 1.1. I agree it should be changed as long as we are ok with the backwards incompatibility

istio-control/istio-autoinject/files/injection-template.yaml
```diff
@@ -11,10 +11,9 @@ template: |
   {{- else }}
     image: "{{ .Values.global.hub }}/{{ .Values.global.proxy_init.image }}:{{ .Values.global.tag }}"
   {{- end }}
-    command:
-    - "/usr/local/bin/istio-iptables.sh"
+    args:
```
costinm: It's intentional - command is better, allows specifying the script. We may run proxy image as proxy_init to improve startup.

istio-control/istio-autoinject/files/injection-template.yaml
```diff
@@ -85,9 +87,9 @@ template: |
     - --domain
     - $(POD_NAMESPACE).svc.{{ .Values.global.proxy.clusterDomain }}
     - --configPath
-    - "/etc/istio/proxy"
+    - "{{ .ProxyConfig.ConfigPath }}"
```
costinm: I don't think either are needed or tested, was planning to deprecate the fields.

istio-control/istio-autoinject/files/injection-template.yaml
```diff
@@ -115,10 +117,12 @@ template: |
     - --datadogAgentAddress
     - "{{ .ProxyConfig.GetTracing.GetDatadog.GetAddress }}"
   {{- end }}
-    - --proxyLogLevel
-    - "{{ annotation .ObjectMeta `sidecar.istio.io/debug` .Values.global.proxy.logLevel}}"
-    - --log_output_level
-    - '{{ .Values.debug }}'
+  {{- if .Values.global.proxy.logLevel }}
```
costinm: This removes the ability to override the log level per pod - which is quite useful.

istio-control/istio-autoinject/files/injection-template.yaml
```diff
@@ -115,10 +117,12 @@ template: |
...
+    - --proxyLogLevel={{ .Values.global.proxy.logLevel }}
+  {{- end}}
+  {{- if .Values.global.proxy.componentLogLevel }}
```
costinm: I would follow the same pattern - log level at a global mesh level is not very useful.

istio-control/istio-autoinject/files/injection-template.yaml
howardjohn: Can we make this `sidecar.istio.io/logLevel` then instead of `debug`?

istio-control/istio-autoinject/files/injection-template.yaml
howardjohn: Honestly not really sure why you would want to change the log level for a single pod. If you are just debugging you can change it at runtime in Envoy.

istio-control/istio-autoinject/values.yaml
```diff
@@ -4,52 +4,14 @@ sidecarInjectorWebhook:
   # retry.
   replicaCount: 1
 
+  image: sidecar_injector
```
costinm: Let's remove this - default is provided, not something we expect users to set. 
Not clear why it was even added.

howardjohn: I think its so you can provide a full image like `image: myrepo/injector:tag`. We need to change this in pilot (maybe others) too if we remove this so they are consistent

global.yaml
costinm: I mean: global.yaml file should have only the global settings - there is a top-level global: 

We want to keep 1.1 compat - if 1.1 used --set istio_cni.enabled=true, it should keep working.
That means anything using istio_cni (injector ?) should have that in its values.yaml

istio-control/istio-autoinject/files/injection-template.yaml
costinm: sure. 

As a general rule - we want the annotations to match the keys used in values.yaml, and eventually to generalize so anything in values.yaml (== mesh config) can be patched by local annotations.

global.yaml
howardjohn: That makes sense, thanks

istio-control/istio-autoinject/files/injection-template.yaml
costinm: Runtime change implies kubectl exec to the machine, running some commands - and if envoy crashes or restarts doing it again. 

I don't want to do it at pod level - but deployment level. For example the staging or test workloads will have it set to debug, while prod to warn.  

The global setting for any control plane that has prod jobs can really only be 'debug off' - I wouldn't mind removing the global and only supporting workload-specific settings.

This happens to be a prefect example for the new patching API and kustomize.

istio-control/istio-autoinject/values.yaml
costinm: Eventually we want to move to kustomize syntax.
```
images:
  - name: postgres
    newName: my-registry/my-postgres
    newTag: v1
  - name: nginx
    newTag: 1.8.0
  - name: my-demo-app
    newName: my-app
  - name: alpine
    digest: sha256:24a0c4b4a4c0eb97a1aabb8e29f18e917d05abfe1b7a7c07857230879ce7d3d3

```

costinm: Which implies the image 'base' name will be fixed in the templates, and user can override it using kustomize. 
We can keep the template and allow image to override, for backward compat - but a missing image in values.yaml should be supported, and we can start deprecating it.

PR #96, opened 2019-05-01
Fix mergify rules

howardjohn:
The circle rules need to match the title of the test. Also added some
extra stuff we are using in docs repo, to prevent accidental merges.


howardjohn:
This will be enabled as soon as we turn this on, it wasn't working before because the check for circleci passing was wrong so it never passed

PR #83, opened 2019-04-29
Isolate components in different namespaces.

morvencao:
Resolve: https://github.com/istio-ecosystem/istio-installer/issues/82

PR #98, opened 2019-05-02
update tracing namespace.

morvencao:
Follow up PR for https://github.com/istio/installer/pull/83 to update tracing namespace.

Tracing component(zipkin or jaeger), if deployed, should be by default located in istioTelemetry namespace.


costinm:
Could you pull master ?

morvencao:
That's right, we already have `telemetryNamespace` in `global.yaml` https://github.com/istio/installer/blob/master/global.yaml#L25

douglas-reid:
@costinm I see no reason why tracing _must_ be in the telemetry namespace. I can see that being managed independently.  And, once `istio-telemetry` disappears into the mist, it should be a tracing specific namespace, imho.

This PR is probably fine for the interim, however.

PR #101, opened 2019-05-02
More kustomization for demo and easy install

costinm:
- add an 'istio-testing' base that will install the demo in istio-testing namespace
- add a kustomize base for crds and ingress ( still needs makefile changes to generate the default ingress)
For 1.2 we can probably allow at least additional gateways to be deployed 'the easy way' via kubectl apply and kustomize.


test/demo/istio-testing/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
+# Installs istio-demo profile in istio-testing environment
+# TODO: override as much as possible
```
howardjohn: What are we overriding? To get the demo values (resource settings, enabled, etc)?

if so shouldnt we just generate those using helm values?

gateways/istio-ingress/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - istio-ingress.yaml
```
howardjohn: This doesn't seem to be a file that exists?

test/demo/istio-testing/kustomization.yaml
costinm: No, we do have the right demo values in test/demo - this is to test how much we can customize the demo with only kustomize.

The 'base' demo is using istio-system - and if we're lucky we may be able to use it as an in-place upgrade from 1.1 ( seems doable - not sure if worth it ).

This is a profile that installs in istio-testing, and will test the limits of kustomization ( == override as much as we can )

gateways/istio-ingress/kustomization.yaml
costinm: fixed - I still don't know if checking in generated files is right, but golang gen files and vendor are doing it so we may be ok for 1.2

costinm:
The main challenge remains citadel in istio-system - still trying to find how to properly configure it

gateways/istio-ingress/kustomization.yaml
howardjohn: I think its ok, as long as its deterministic. There was some stuff different when I ran it vs yours - some stuff has `costin` in it at the very least and I think some may not have consistent ordering.

PR #102, opened 2019-05-02
Sync prometheus with master

howardjohn:


costinm:
Should we just remove this - and use Doug's PR to us prom operator ? 
@douglas-reid @mandarjog

howardjohn:
I think the plan was to support both - our "demo" Prometheus and bring your own operator. But up to telemetry experts

douglas-reid:
@costinm @howardjohn I think for the 1.3 timeframe, we can think about completely removing the demo prom stuff... but since the operator stuff hasn't been put through its paces yet, I don't know that I'm comfortable with getting rid of the legacy prom operator stuff yet.

Does that match with the env WG pattern?

PR #103, opened 2019-05-02
First pass at getting things to work on a Mac

ericvn:
- Docker image creation
  - tests


costinm:
I suspect the tests won't pass until a new image is built - can you add back the mount for docker (maybe with an option ), we can remove it after I rebuild the image.

ericvn:
Looking at some of the failures above, they failed with:
```
Error: could not list clusters: failed to list nodes: exec: "docker": executable file not found in $PATH
```
These are because the current istionightly/kind:v1.14.1-1 does not have the `docker` executable as that change is part of the new istionightly/kind:v1.14.1-1 using this Dockerfile.

PR #105, opened 2019-05-03
Sync grafana templates with master

howardjohn:


istio-telemetry/grafana/templates/configmap.yaml
```diff
@@ -27,9 +26,9 @@ data:
         timeInterval: 5s
       editable: true
 
-{{- if .Values.grafana.datasources }}
+{{- if $datasources.datasources }}
     -
-  {{- range $key, $value := .Values.grafana.datasources }}
```
howardjohn: @douglas-reid what do you think of how this works? The problem is we need to be able to set the prometheusNamespace now, which is why we have the hardcoded source above.

PR #111, opened 2019-05-05
Add an option to use a self-signed cert for autoinject webhook

philrud:
Introduces a 'sidecarInjectorWebhook.selfSigned' option to the 'istio-autoinject'
component that allows to generate a self-signed CA and issue a long-lived
certificate during template rendering that can be used to authorize a webhook
workload.

### DIFF
--- a/istio-control/istio-autoinject/values.yaml
+++ b/istio-control/istio-autoinject/values.yaml
@@ -16,6 +16,11 @@ sidecarInjectorWebhook:
   # even when mTLS is enabled.
   rewriteAppHTTPProbe: false
 
+  # If true, a self-signed CA will created in order to issue a certificate that
+  # will be used to authenticate the workload respondible for handling
+  # the sidecar-injector webhook.
+  selfSigned: false
+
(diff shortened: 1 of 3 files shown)


istio-control/istio-autoinject/values.yaml
```diff
@@ -16,6 +16,11 @@ sidecarInjectorWebhook:
   # even when mTLS is enabled.
   rewriteAppHTTPProbe: false
 
+  # If true, a self-signed CA will created in order to issue a certificate that
```
costinm: Let's add a comment here that short term this is implemented using helm cert generation, and will migrate into istioctl.

Also mention that the cert private will be included in the helm template generation - so should only be used with iop install ( and make sure the iop doesn't save the template gen or removes it after ).

costinm: Can you add a test ( in test/*.mk, to one of the targets ) ? Just to verify that a custom namespace can be used to install an injector, and maybe used.

PR #112, opened 2019-05-06
Remove mount of docker as it's part of new kind image

ericvn:


howardjohn:
You can remove the label when all tests are passing, we need to update mergify to require all the tests now that they should all pass.

But looks like they are failing from missing docker

ericvn:
It looks like the `istionightly/kind:v1.14.1-1` might not have been after https://github.com/istio/installer/pull/103. Checking that now.

howardjohn:
I think it is manually built by costin

ericvn:
Looks like it wasn't built again after #103 was merged. Will work with costing to get that pushed.

Locally built image:
```
# ls /usr/local/bin
ci2gubernator  clean-install  crictl  ctr  docker  entrypoint  go-junit-report	helm  istioctl	kind  repo
```
Currently checked in image:
```
# ls /usr/local/bin
ci2gubernator  clean-install  crictl  entrypoint  go-junit-report  helm  istioctl  kind  repo
```

PR #114, opened 2019-05-06
Require all checks for mergify

howardjohn:


.mergify.yml
```diff
@@ -3,6 +3,9 @@ pull_request_rules:
     conditions:
       - "status-success=ci/circleci: build"
       - "status-success=ci/circleci: noauth"
+      - "status-success=ci/circleci: install-minikube"
+      - "status-success=ci/circleci: integration-old-kind"
+      - "status-success=ci/circleci: integration-presubmit-kind"
```
linsun: @howardjohn is this test stable?  your PR failed on it.

howardjohn: It was stable, but something change in istio/istio causing the failure, I am looking into it now.

costinm:
We really need a test in istio/istio that clones istio/installers and runs one of those tests.

linsun:
ok to merge this now?  assume you'll get all failed 2 tests passing

howardjohn:
@linsun rebasing to make sure test pass, ill merge when they do

PR #115, opened 2019-05-06
Fix overridden galley fixes

howardjohn:
I fixed this issues in my pr to sync galley to master, but accidentally overrode them later, so they got missed.

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -44,9 +44,6 @@ global:
   logging:
     level: "default:info"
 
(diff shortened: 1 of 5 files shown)


howardjohn:
Testing if mergify requires owner approval or any approval, Eric may or may not actually approve.

Looks like it accepts any approval

costinm:
Seems to be failing a policy / redis test ? Can we disable if it's not supported in this environment ?

PR #107, opened 2019-05-03
Improve integration tests

howardjohn:
Run in parallel, skip tests that require custom setup.

Close to being able to run all of them (not just the small presubmit
subset).

PR #123, opened 2019-05-07
Isolate components in different namespaces.

morvencao:
Resolve: #82
Context: https://github.com/istio/installer/pull/83

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -210,7 +210,7 @@ jobs:
 
   # Verify a-la-carte with no certificates ( smallest possible install, and cases without certs/mtls)
   noauth:
-    <<: *integrationRemote
+    <<: *integrationRemoteOneNamespace
 
     # Can run multiple images, first is primary, all use localhost
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,11 +23,11 @@ integrationRemoteOneNamespace: &integrationRemoteOneNamespace
         ONE_NAMESPACE: 1
         TMPDIR: /go/tmp
 
-# remote docker environment for isolated namespaces.
+# remote docker environment for one namespace. Appears to be faster
 integrationRemoteIsolatedNamespaces: &integrationRemoteIsolatedNamespaces
   working_directory: /go/src/istio.io/installer
   docker:
-    - image: istio/ci:go1.11-k8s1.10.4-helm2.7.2-minikube0.25
+    - image: istionightly/ci:2019-05-09
       environment:
         GOPATH: /go
         SKIP_CLEANUP: 1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,16 +12,26 @@ integrationDefaults: &integrationDefaults
     - KUBECONFIG: /go/out/minikube.conf
     - TEST_ENV: minikube-none
 
-# remote docker environment. Appears to be faster
-integrationRemote: &integrationRemote
+# remote docker environment for one namespace. Appears to be faster
+integrationRemoteOneNamespace: &integrationRemoteOneNamespace
   working_directory: /go/src/istio.io/installer
   docker:
     - image: istionightly/ci:2019-05-09
       environment:
         GOPATH: /go
         SKIP_CLEANUP: 1
+        ONE_NAMESPACE: 1
         TMPDIR: /go/tmp
 
+# remote docker environment for isolated namespaces.
+integrationRemoteIsolatedNamespaces: &integrationRemoteIsolatedNamespaces
+  working_directory: /go/src/istio.io/installer
+  docker:
+    - image: istio/ci:go1.11-k8s1.10.4-helm2.7.2-minikube0.25
+      environment:
+        GOPATH: /go
+        SKIP_CLEANUP: 1
+        TMPDIR: /go/tmp
 
 # Common procedure to initialize working directory
 initWorkingDir: &initWorkingDir
@@ -41,8 +51,29 @@ initWorkingDir: &initWorkingDir
 
 jobs:
 
-  # Build all the templates, verify the templates work.
-  build:
+  # Build all the templates in one namespace, verify the templates work.
+  build_one_namespace:
+    working_directory: /go/src/istio.io/installer
+    docker:
+      # Image with KIND installed, based on Istio Kind node-image
+      # Should have all tools we need, recreated when we rev the tool versions.
+      - image: istio/ci:go1.11-k8s1.10.4-helm2.7.2-minikube0.25
+        environment:
+          GOPATH: /go
+          TMPDIR: /go/tmp
+          ONE_NAMESPACE: 1
+
+    # Can run multiple images, first is primary, all use localhost
+    steps:
+      - checkout
+      - run:
+          command: make run-build
+          # TODO: make lint
+      - store_artifacts:
+          path: /go/out/
+
+  # Build the templates in isolated namespaces, verify the templates work.
+  build_isolated_namespaces:
     working_directory: /go/src/istio.io/installer
     docker:
       # Image with KIND installed, based on Istio Kind node-image
@@ -65,8 +96,8 @@ jobs:
   # Pipeline running KIND and installer in the remote-docker environment
   # Runs the 'integration.presubmit' tests against an environment using new installer with components in same
   # namespace.
-  integration-presubmit-kind:
-    <<: *integrationRemote
+  integration-presubmit-one-namespace-kind:
+    <<: *integrationRemoteOneNamespace
 
     # Can run multiple images, first is primary, all use localhost
     steps:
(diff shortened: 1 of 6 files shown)


morvencao:
Just found two issues that block the CI: https://github.com/istio/installer/issues/124 and https://github.com/istio/installer/issues/125

morvencao:
The reasons for three failed CI cases:
1. For `integration-old-isolated-namespaces-kind`: we need to enhance Istio e2e test framework to add isolated namespaces support, see: https://github.com/istio/istio/blob/master/tests/e2e/framework/kubernetes.go#L94

2. For `integration-presubmit-isolated-namespaces-kind` and `integration-presubmit-one-namespace-kind` cases, they both failed due to the pod permission issue of test chart `stable/redis`; it's probably not caused by istio, because I have try to deploy it without sidecar in the same cluster, the pods still can't be started.

@howardjohn @costinm Shall we consider merging this ?

howardjohn:
I have a fix for the redis stuff in https://github.com/istio/installer/pull/107. (not a real fix, but it disables the test).

As for the old one, we may just want to disable it for now - I don't think there are any plans to invest effort into changing the old e2e tests, so it may never be fixed?

morvencao:
@howardjohn Great, thanks.

If e2e test framework stays unchanged, then we should remove the `integration-old-isolated-namespaces-kind` case.

.circleci/config.yml
```diff
@@ -12,16 +12,26 @@ integrationDefaults: &integrationDefaults
     - KUBECONFIG: /go/out/minikube.conf
     - TEST_ENV: minikube-none
 
-# remote docker environment. Appears to be faster
-integrationRemote: &integrationRemote
+# remote docker environment for one namespace. Appears to be faster
+integrationRemoteOneNamespace: &integrationRemoteOneNamespace
```
costinm: The real solution is to make sure the tests are compatible with multiple namespaces. 
How about we keep the 'one namespace' option only for demo ? I don't think we need more tests for one namespace
since it'll not be recommended.

.circleci/config.yml
```diff
@@ -41,8 +51,29 @@ initWorkingDir: &initWorkingDir
 
 jobs:
 
-  # Build all the templates, verify the templates work.
-  build:
+  # Build all the templates in one namespace, verify the templates work.
```
costinm: That's now done by the 'build' target - it builds both the demo ( one ns ) and few other templates.
This is also used for 'kubectl apply -k"

.circleci/config.yml
morvencao: There are some gaps that will break `isolated-namespaces` case if we want to reuse the e2e test framework, eg. `old-integration`. 
Fortunately, we get all current test cases passed.

morvencao:
@howardjohn Thanks for https://github.com/istio/installer/pull/107, finally we get the tests passed.

Makefile
```diff
@@ -137,15 +152,27 @@ build:
...
+else
+	docker exec -e KUBECONFIG=/etc/kubernetes/admin.conf \
 		${KIND_CLUSTER}-control-plane \
 		bash -c "cd ${GOPATH}/src/istio.io/installer; make ${TARGET}"
+endif
 
 # Runs the test in docker. Will exec into KIND and run "make $TEST_TARGET" (default: run-all-tests)
 docker-run-test:
-	docker exec -e KUBECONFIG=/etc/kubernetes/admin.conf  \
+ifeq ($(ONE_NAMESPACE), 1)
```
howardjohn: I wonder if we can just pass down all env variables so we don't need special cases like this? Not needed for this PR but something to keep in mind for the future.

nit: can we just do ONE NAMESPACE=$(ONE NAMESPACE) instead of if statement?

test/tests.mk
```diff
@@ -107,7 +107,7 @@ run-test.integration.kube:
 
 	kubectl -n default apply -f test/k8s/mtls_permissive.yaml
 	kubectl -n default apply -f test/k8s/sidecar-local.yaml
-
+
```
howardjohn: nit: extra space

morvencao: done

Makefile
morvencao: nice catch.

.circleci/config.yml
costinm: I would still like to make it clear this is a 'demo' profile only - to avoid confusion. I think it is very important for security and operations to move to separate namespace and stop supporting 'one namespace'.

.circleci/config.yml
```diff
@@ -246,18 +352,24 @@ workflows:
...
-      - noauth
+      - noauth_one_namespace
+      - noauth_isolated_namespaces
 
   all:
     jobs:
-      - build
-      - integration-presubmit-kind
+      - build_one_namespace
+      - build_isolated_namespaces
```
costinm: Probably a single build is sufficient - it should build all combinations ( it does build demo and others )

morvencao:
@howardjohn Thanks, raise a separate PR: https://github.com/istio/installer/pull/149, PTAL

.circleci/config.yml
```diff
@@ -159,8 +207,39 @@ jobs:
...
+      - run:
+          when: always
+          command: make kind-logs
+      - store_test_results:
+          path: /go/out/tests
+      - store_artifacts:
+          path: /go/out/logs
+
+  # Verify a-la-carte with no certificates ( smallest possible install, and cases without certs/mtls)
+  noauth_isolated_namespaces:
```
costinm: I don't think we need this - the purpose of noauth is to verify the config used by knative and few others, as well as the case where user deploys istio gateway in their own namespace (next to the app). 

It also serves to test the case where dedicated pilots are deployed for gateway only, to minimize risks and improve 
segmentation.

If there is user demand for pilot in 1 ns, and ingress in a different ns - we can add it, but I don't want to add 
profiles without a clear demand, there are more useful configs we can test.

.circleci/config.yml
```diff
@@ -239,17 +318,21 @@ workflows:
...
       - build
-      - integration-presubmit-kind
+      - integration-presubmit-one-namespace-kind
+      - integration-presubmit-isolated-namespaces-kind
       - integration-old-kind
       - install-minikube
       - install-demo
-      - noauth
+      - noauth_one_namespace
+      - noauth_isolated_namespaces
```
costinm: Also: if you want to test this case, no need to create a new job - each job can test different scenarios in different namespaces. For local testing - we do need to keep distinct cluster for 'noauth' type of tests because we'll not 
have citadel and need to test that everything works in the absence of certs. But multiple install profiles that 
are in this situation can use the same cluster - we also want to verify they don't interfere with each other. 

( we are still short on resources - each job takes ~10.15 min I think - if you really want this profile as a separate job, maybe add it to nightly ?)

.circleci/config.yml
morvencao: Thanks, got it!

test/tests.mk
```diff
@@ -88,13 +88,13 @@ INT_FLAGS ?= \
...
 	--istio.test.nocleanup \
 	--istio.test.kube.deploy=false \
-	--istio.test.kube.systemNamespace ${ISTIO_NS} \
-	--istio.test.kube.istioNamespace ${ISTIO_NS} \
-	--istio.test.kube.configNamespace ${ISTIO_NS} \
-	--istio.test.kube.telemetryNamespace ${ISTIO_NS} \
-	--istio.test.kube.policyNamespace ${ISTIO_NS} \
-	--istio.test.kube.ingressNamespace ${ISTIO_NS} \
-	--istio.test.kube.egressNamespace ${ISTIO_NS} \
+	--istio.test.kube.systemNamespace ${ISTIO_CONTROL_NS} \
```
c0d1ngm0nk3y: not `{ISTIO_SYSTEM_NS}`?

test/tests.mk
```diff
@@ -121,16 +121,18 @@ run-test.integration.kube.presubmit:
...
+	 ISTIO_ENV=${ISTIO_CONTROL_NS} bin/iop mysqlplain mysqlplain ${BASE}/test/mysql ${IOP_OPTS} --set mtls=false --set Name=plain
 
 # This test currently only validates the correct config generation and install in API server.
 # When prom operator config moves out of alpha, this should be incorporated in the other tests
 # and removed.
 run-prometheus-operator-config-test: PROM_OPTS="--set prometheus.createPrometheusResource=true"
 run-prometheus-operator-config-test: install-prometheus-operator install-prometheus-operator-config
-	if [ "$$(kubectl -n ${ISTIO_NS} get servicemonitors -o name | wc -l)" -ne "7" ]; then echo "Failure to find ServiceMonitor resouces!"; return 1; fi
-	kubectl -n ${ISTIO_NS} wait deploy/prometheus --for=condition=available --timeout=${WAIT_TIMEOUT}
+	if [ "$$(kubectl -n ${ISTIO_CONTROL_NS} get servicemonitors -o name | wc -l)" -ne "7" ]; then echo "Failure to find ServiceMonitor resouces!"; return 1; fi
```
c0d1ngm0nk3y: typo `resouces`

test/tests.mk
```diff
@@ -121,16 +121,18 @@ run-test.integration.kube.presubmit:
...
+	# kubectl wait is problematic, as the pod may not exist before the command is issued.
+	until timeout ${WAIT_TIMEOUT} kubectl -n ${ISTIO_CONTROL_NS} get pod/prometheus-prometheus-0; do echo "Waiting for pods to be created..."; done
+	kubectl -n ${ISTIO_CONTROL_NS} wait pod/prometheus-prometheus-0 --for=condition=Ready --timeout=${WAIT_TIMEOUT}
```
c0d1ngm0nk3y: `prometheus` is always in `istio-control`?

costinm: No, certainly not - should be in istio-telemetry or its own namespace ( or whatever the user wants - but we don't recommend keeping it under control due to risks )

costinm:
Rebase please.

test/tests.mk
morvencao: discovery service is installed in `${ISTIO_CONTROL_NS}`

Issue #125, opened 2019-05-07
Insufficient Resource for `install-demo` test


morvencao:
In [CircleCI install-demo test](https://circleci.com/gh/istio/installer/1446#queue-placeholder/containers/0) env, I found that `istio-telemetry` pod can't be started because of Insufficient Resource in kind cluster:

```
root@test-control-plane:/# kubectl -n istio-system get pod
NAME                                     READY   STATUS    RESTARTS   AGE
egressgateway-696fd5745f-tpd6s           1/1     Running   0          84s
grafana-669d58766-kgj8f                  1/1     Running   0          84s
ingressgateway-f7748bd97-bxhdt           0/1     Running   0          84s
istio-citadel11-5ff58996dd-mxlfg         1/1     Running   0          84s
istio-galley-7f6b46c947-5w8mz            2/2     Running   0          84s
istio-pilot-67857ff455-5hcr4             1/2     Running   0          84s
istio-sidecar-injector-878bb7c96-z7n8c   1/1     Running   0          84s
istio-telemetry-bc45c5dd5-4878j          0/2     Pending   0          84s
prometheus-58cf46cc44-wchmx              1/1     Running   0          84s
root@test-control-plane:/# kubectl -n istio-system describe pod istio-telemetry-bc45c5dd5-4878j
Name:               istio-telemetry-bc45c5dd5-4878j
Namespace:          istio-system
Priority:           0
PriorityClassName:  <none>
Node:               <none>
...
Events:
  Type     Reason            Age                From               Message
  ----     ------            ----               ----               -------
  Warning  FailedScheduling  27s (x3 over 96s)  default-scheduler  0/1 nodes are available: 1 Insufficient cpu.
```

Resulting in `ci/circleci: install-demo` failure.

/cc @howardjohn @costinm

howardjohn:
Fixed in #127, thanks for figuring this out

PR #127, opened 2019-05-07
Use demo values for demo

howardjohn:
Hopefully fix https://github.com/istio/installer/issues/125

Most of changes are from updating the demo yaml, the real change is the values.yaml

### DIFF
--- a/test/demo/k8s.yaml
+++ b/test/demo/k8s.yaml
@@ -12,7 +12,7 @@ metadata:
     istio: sidecar-injector
 data:
   values: |-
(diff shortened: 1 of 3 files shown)


test/demo/k8s.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
...
..."requests":{"cpu":"10m","memory":"40Mi"}},"statusPort":15020,"tracer":"zipkin"},"proxy_init":{"image":"proxy_init"},"sds":{"enabled":false,"udsPath":"","useNormalJwt":false,"useTrustworthyJwt":false},"sidecarInjectorWebhook":{"enabled":true,"rewriteAppHTTPProbe":false},"tag":"nightly-master","telemetryNamespace":"istio-system","tracer":{"datadog":{"address":"$(HOST_IP):8126"},"lightstep":{"accessToken":"","address":"","cacertPath":"","secure":true},"zipkin":{"address":""}},"trustDomain":"","useMCP":true},"grafana":{"enabled":true},"istio_cni":{"enabled":false},"kiali":{"createDemoSecret":true,"enabled":true},"mixer":{"adapters":{"stdio":{"enabled":true}},"policy":{"enabled":true,"resources":{"limits":{"cpu":"100m","memory":"100Mi"},"requests":{"cpu":"10m","memory":"100Mi"}}},"telemetry":{"enabled":true,"resources":{"limits":{"cpu":"100m","memory":"100Mi"},"requests":{"cpu":"50m","memory":"100Mi"}}}},"pilot":{"resources":{"limits":{"cpu":"100m","memory":"200Mi"},"requests":{"cpu":"10m","memory":"100Mi"}},"traceSampling":100},"sidecarInjectorWebhook":{"enableNamespacesByDefault":false,"image":"sidecar_injector","replicaCount":1,"rewriteAppHTTPProbe":false},"tracing":{"enabled":true}}
```
linsun: is this generated?  this is a little harder to review

howardjohn: Yes the whole k8s.yaml file is genreated with `make run-build-demo`. In istio/istio we didn't check these in, which maybe we dont need to, but it does allow users to `kubectl apply -k github.com/istio/install/demo/k8s.yaml`.

Specifically, this part which is hard to review, is the values file serialized, so you can review the values.yaml.

But I agree, we probably need a better solution

linsun: ok, i'll ignore reviewing them.  certainly not the normal k8s yaml file i see.

linsun:
@howardjohn should we make this demo test a required test now?

howardjohn:
@linsun We should make all of the tests required I think. I had a PR here but its failing (and the demo one is new), so once we get all them passing we should do that. Things are moving pretty quick in this repo so it is hard to keep up with all the new tests, etc

Issue #116, opened 2019-05-06
sync pod affinity stuff to installer from master and release-1.1


linsun:
@morvencao I don't think your change related to pod affinity config in the new installer.

Could you open a PR to merge them in?  Didn't want to miss it!

morvencao:
Close this since we have merged https://github.com/istio/installer/pull/128.

PR #128, opened 2019-05-07
add pod anti affinity.

morvencao:
Resolve: https://github.com/istio/installer/issues/116


gateways/istio-ingress/values.yaml
```diff
@@ -109,4 +109,23 @@ gateways:
...
+    # For example:
+    # podAntiAffinityLabelSelector:
+    # - key: security
+    #   operator: In
+    #   values: S1,S2
+    #   topologyKey: "kubernetes.io/hostname"
+    # This pod anti-affinity rule says that the pod requires not to be scheduled
+    # onto a node if that node is already running a pod with label having key
+    # “security” and value “S1”.
+    podAntiAffinityLabelSelector: {}
```
costinm: What is the effect of empty selector - will it trigger the if ? Don't remember of {{ if {} }} is true of false...

morvencao: Will not trigger if for empty selector.

Issue #113, opened 2019-05-06
Port other chart charts


howardjohn:
We are currently missing:
* kiali
* tracing
* nodeagent
* istiocoredns
* certmanager

richardwxn:
FYI, I have a PR open for kiali #104 and I am working on the tracing part as well.

PR #132, opened 2019-05-07
Import tracing related templates

richardwxn:
Work as part of #113

### DIFF
--- a/istio-telemetry/tracing/templates/deployment-jaeger.yaml
+++ b/istio-telemetry/tracing/templates/deployment-jaeger.yaml
@@ -1,0 +1,81 @@
+{{ if eq .Values.tracing.provider "jaeger" }}
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: istio-tracing
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: jaeger
+    release: {{ .Release.Name }}
+spec:
+  selector:
+    matchLabels:
+      app: jaeger
+  template:
+    metadata:
+      labels:
+        app: jaeger
+        release: {{ .Release.Name }}
+      annotations:
+        sidecar.istio.io/inject: "false"
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "16686"
+{{- if .Values.tracing.contextPath }}
+        prometheus.io/path: "{{ .Values.tracing.contextPath }}/metrics"
+{{- else }}
+        prometheus.io/path: "/{{ .Values.tracing.provider }}/metrics"
+{{- end }}
+    spec:
+{{- if .Values.global.priorityClassName }}
+      priorityClassName: "{{ .Values.global.priorityClassName }}"
+{{- end }}
+{{- if .Values.global.imagePullSecrets }}
+      imagePullSecrets:
+{{- range .Values.global.imagePullSecrets }}
+        - name: {{ . }}
+{{- end }}
+{{- end }}
+      containers:
+        - name: jaeger
+          image: "{{ .Values.tracing.jaeger.hub }}/all-in-one:{{ .Values.tracing.jaeger.tag }}"
+          imagePullPolicy: {{ .Values.global.imagePullPolicy }}
+          ports:
+            - containerPort: 9411
+            - containerPort: 16686
+            - containerPort: 5775
+              protocol: UDP
+            - containerPort: 6831
+              protocol: UDP
+            - containerPort: 6832
+              protocol: UDP
+          env:
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                apiVersion: v1
+                fieldPath: metadata.namespace
+          - name: COLLECTOR_ZIPKIN_HTTP_PORT
+            value: "9411"
+          - name: MEMORY_MAX_TRACES
+            value: "{{ .Values.tracing.jaeger.memory.max_traces }}"
+          - name: QUERY_BASE_PATH
+            value: {{ if .Values.tracing.contextPath }} {{ .Values.tracing.contextPath }} {{ else }} /{{ .Values.tracing.provider }} {{ end }}
+          livenessProbe:
+            httpGet:
+              path: /
+              port: 16686
+          readinessProbe:
+            httpGet:
+              path: /
+              port: 16686
+          resources:
+{{- if .Values.tracing.jaeger.resources }}
+{{ toYaml .Values.tracing.jaeger.resources | indent 12 }}
+{{- else }}
+{{ toYaml .Values.global.defaultResources | indent 12 }}
+{{- end }}
+      affinity:
+      {{- include "nodeaffinity" . | indent 6 }}
+      {{- include "podAntiAffinity" . | indent 6 }}
+{{ end }}
--- a/istio-telemetry/tracing/templates/service.yaml
+++ b/istio-telemetry/tracing/templates/service.yaml
@@ -1,0 +1,41 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: zipkin
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Values.tracing.provider }}
+    release: {{ .Release.Name }}
+spec:
+  type: {{ .Values.tracing.service.type }}
+  ports:
+    - port: {{ .Values.tracing.service.externalPort }}
+      targetPort: 9411
+      protocol: TCP
+      name: {{ .Values.tracing.service.name }}
+  selector:
+    app: {{ .Values.tracing.provider }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: tracing
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    {{- range $key, $val := .Values.tracing.service.annotations }}
+    {{ $key }}: {{ $val | quote }}
+    {{- end }}
+  labels:
+    app: {{ .Values.tracing.provider }}
+    release: {{ .Release.Name }}
+spec:
+  ports:
+    - name: http-query
+      port: 80
+      protocol: TCP
+{{ if eq .Values.tracing.provider "jaeger" }}
+      targetPort: 16686
+{{ else }}
+      targetPort: 9411
+{{ end}}
+  selector:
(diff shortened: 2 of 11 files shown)


istio-telemetry/tracing/templates/service.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: v1
+kind: List
```
howardjohn: I know this is in master, but this is weird? Why do we have List?

istio-telemetry/tracing/templates/service.yaml
```diff
@@ -0,0 +1,50 @@
...
+      labels:
+        app: {{ .Values.tracing.provider }}
+        release: {{ .Release.Name }}
+    spec:
+      ports:
+        - name: http-query
+          port: 80
+          protocol: TCP
+  {{ if eq .Values.tracing.provider "jaeger" }}
+targetPort: 16686
```
howardjohn: Indent seems to have gotten messed up?

richardwxn: yes good catch, fixed it

douglas-reid:
@howardjohn I think the true path forward is to switch to operator controlled installs.  In the meantime, it is probably best to start with just matching the current way, and then adapting over time -- but I'll defer to @objectiser, who has spent a lot more time and focus on the install artifacts for tracing.

howardjohn:
I guess one difference in this compared to gateways and mixer which we split is that you would never have both tracers installed (right?) so maybe it is less useful

istio-telemetry/tracing/templates/service.yaml
richardwxn: Actually I am not sure about why the second Service named tracing in the list exist? might need some confirmation from the owner.

objectiser: @howardjohn Pending merge: https://github.com/istio/istio/pull/12669

@richardwxn The `tracing` service provides a provider independent service that can be used to access the tracing UI - e.g. as referenced in https://istio.io/docs/tasks/telemetry/gateways/

istio-telemetry/tracing/values.yaml
```diff
@@ -0,0 +1,77 @@
+#
+# addon tracing configuration
+#
+tracing:
+  enabled: false
+
+  provider: zipkin
```
objectiser: Is the default provider being changed for a particular reason?

objectiser:
Agree with @douglas-reid - best to copy over current approach and then move to operators where supported. Once this is merged, I'll create a PR along the lines of the prometheus operator PR for Jaeger.

@howardjohn Yes only one tracer should be installed - as selected by the `tracing.provider` switch.

istio-telemetry/tracing/templates/service.yaml
howardjohn: Yeah I think we should copy the changes from 12669 here, no reason to have this as a List. Thanks @objectiser

istio-telemetry/tracing/values.yaml
howardjohn: I have never even used tracing, so not an expert on it by any means, but why do we have tracing.provider=jaeger and global.proxy.tracer=zipkins? what is even the difference? We have so many different tracing configs..

objectiser: I know it can be confusing. The `tracing.provider` refers to the backend tracing system. The `global.proxy.tracer` refers to the tracing client being used by the proxy - i.e. could be lightstep, zipkin, etc.

Jaeger can consume the zipkin data format, hence the tracer can be zipkin, but the tracing.provider can be jaeger.

howardjohn: Thanks for clearing that up. If that is the case then it looks like this was not intentionally changed, we should keep in line with current master (Jaeger)?

objectiser: +1

istio-telemetry/tracing/templates/deployment-jaeger.yaml
```diff
@@ -0,0 +1,81 @@
+{{ if eq .Values.tracing.provider "jaeger" }}
```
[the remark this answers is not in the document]
douglas-reid: My suggestion continues to be to take the existing templates and consider tracing as a unit for now -- just for simplicity's sake. We could then immediately start picking apart what we want.

However, we _could_ break everything apart. That might lead to a bit of duplication and overlap, as both backends share the same frontend service spec, but perhaps that's not a big deal. (I don't believe that having to install both "istio-tracing" and an implementation deployment is a great user experience -- seems like installing a single installation other should do everything required).

@objectiser has an operator-based install for `jaeger` in dev and is basically waiting for the dust to settle here.

If we chose to break this apart, I think this means the structure looks like:

```
istio-telemetry/
|   jaeger/
|   jaeger-operator/
|   zipkin/
```

Wherein all three installs include the definition of the "zipkin" and "tracing" services.

Is that a shared understanding?  Or are there better ideas?

istio-telemetry/tracing/values.yaml
richardwxn: I changed this when I was testing with zipkin locally and forgot to revert it back, change back to jaeger now.

istio-telemetry/tracing/templates/service.yaml
richardwxn: Removed the List part

istio-telemetry/tracing/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
+{{/* vim: set filetype=mustache: */}}
```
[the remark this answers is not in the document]
richardwxn: Can you give an example of it? I thought it is just for syntax purpose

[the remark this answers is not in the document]
howardjohn: Not talking about the filetype stuff, the entire use of things in the _helpers. We should use {{ template "name" }} or what every it is, just use `name: tracing`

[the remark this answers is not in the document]
richardwxn: sure, remove related definition in _helpers. The _helper files for other components need clean up as well.

istio-telemetry/tracing/templates/deployment-jaeger.yaml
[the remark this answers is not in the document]
costinm: Installing both backends - or different versions of each backend - is certainly in scope and goal of the new installer. But in different 'environments'/namespaces. 

Our experience with the old installer is that we almost never go back to fix - and now is the best opportunity to make some backward incompatible changes where it makes sense for the modular installer. 

Given that prometheus for example is moving completely to operator and independent operation - I don't see the problem here.

If we want to start with something simple - we can just pick one of the 2, for example jaeger-operator - get it added by itself, and we can work in parallel on zipkin.

Not sure we even need both jaeger and jaeger-opertor.

morvencao:
We need to include the tracing test case once tracing component is imported, currently it's excluded https://github.com/istio/installer/blob/master/test/tests.mk#L80

/cc @howardjohn

costinm:
Ping ? 
Given the time left - we'll need to merge it as is, but at least get rid of the Ingress ( see Kiali change, we want all ingress controlled in the gateway namespace, where we can get certs, etc). And of course we want to use Gateway for our own components, k8s ingress may or may not be installed.

costinm:
One failing test 
```
serviceaccount/prometheus configured
unable to recognize "STDIN": no matches for kind "Prometheus" in version "monitoring.coreos.com/v1"
unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
make: *** [test/install.mk:151: install-prometheus-operator-config] Error 1
Makefile:146: recipe for target 'docker-run-test' failed
make: *** [docker-run-test] Error 2
```

Maybe merge master ?

costinm:
approve otherwise - but the bot will merge if I click the approve button.

douglas-reid:
> One failing test
> 
> ```
> serviceaccount/prometheus configured
> unable to recognize "STDIN": no matches for kind "Prometheus" in version "monitoring.coreos.com/v1"
> unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
> unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
> unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
> unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
> unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
> unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
> unable to recognize "STDIN": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
> make: *** [test/install.mk:151: install-prometheus-operator-config] Error 1
> Makefile:146: recipe for target 'docker-run-test' failed
> make: *** [docker-run-test] Error 2
> ```
> 
> Maybe merge master ?

This seems to happen when the `until` command doesn't work in waiting for the CRDs to be pushed by the operator. We never quite figured out why what would happen.

richardwxn:
> approve otherwise - but the bot will merge if I click the approve button.

@costinm Thanks for checking. Rerun the test and it pass now

costinm:
Sorry, one more - conflicts.

PR #134, opened 2019-05-07
Sync egressgateway with master

howardjohn:


gateways/istio-egress/templates/_affinity.tpl
```diff
@@ -1,28 +1,35 @@
...
 
-{{- define "nodeAffinityRequiredDuringScheduling" }}
+{{- define "gatewayNodeAffinityRequiredDuringScheduling" }}
       nodeSelectorTerms:
       - matchExpressions:
         - key: beta.kubernetes.io/arch
           operator: In
           values:
-        {{- range $key, $val := .Values.global.arch }}
+        {{- range $key, $val := .root.Values.global.arch }}
```
costinm: what is .root ?

howardjohn: If you look later down when we call it: `      {{- include "gatewaynodeaffinity" (dict "root" $ "nodeSelector" $gateway.nodeSelector) | indent 6 }}`

We pass a `.root`

This template is so complicated I was worried changing it would break it -- probably not a good state to be in. I can try to clean this up.

Ideally i think we would have no affinity stuff, and just make users kustomize it

PR #136, opened 2019-05-08
Updates to get `make test` to work

ericvn:
- Extend default test timeout from 3m to 5m

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -24,7 +24,7 @@ function print_help() {
 # Customizable
 INGRESS_NS=${INGRESS_NS:-istio-ingress}
 
-export WAIT_TIMEOUT=${WAIT_TIMEOUT:-3m}
+export WAIT_TIMEOUT=${WAIT_TIMEOUT:-5m}
 SKIP_CLEANUP=${SKIP_CLEANUP:-0}
 SKIP_SETUP=${SKIP_SETUP:-0}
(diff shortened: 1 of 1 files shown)


ericvn:
Marked as DNM for now as I still need to verify all tests pass.

howardjohn:
Was it timing out in ci or locally? I've only had it time out from image pulls when I am on a bad connection so just wondering if there is a different issue or just that?

ericvn:
Timeout in Booking:
```
+ for depl in ${BOOKINFO_DEPLOYMENTS}
+ kubectl -n bookinfo rollout status deployments details-v1 --timeout=3m
Waiting for deployment "details-v1" rollout to finish: 0 of 1 updated replicas are available...
error: timed out waiting for the condition
make: *** [test/tests.mk:66: run-bookinfo] Error 1
```

ericvn:
I removed the run-prometheus-operator-config-test test from run-all-tests since it fails to run locally,
I don't see where it has been added to circleCI so it's not routinely tested.

ericvn:
Removing the DNM label since with these changes `make test` now works for me locally.

That said, I'm still trying to figure out why the run-prometheus-operator-config-test test is failing and getting it added back in (and added to CircleCI).

ericvn:
Worked with @douglas-reid on https://github.com/istio/installer/pull/137 which should fix the test so will add it back in.

Issue #135, opened 2019-05-08
running make test locally failed


linsun:
with 2 pods not able to reach running: (after excluding the mixer redis testing)

```
root@test-control-plane:/Users/linsun/go/src/istio.io/installer# kubectl describe pod nginx-default-http-backend-84bdb65549-nwwkd  -n simple-strict
  Normal   Pulled     14m (x3 over 16m)   kubelet, test-control-plane  Container image "gcr.io/google_containers/defaultbackend:1.4" already present on machine
  Warning  Unhealthy  10m (x7 over 12m)   kubelet, test-control-plane  (combined from similar events): Liveness probe failed: Get http://10.32.0.24:8080/healthz: read tcp 10.32.0.1:60862->10.32.0.24:8080: read: connection reset by peer
  Warning  BackOff    70s (x39 over 10m)  kubelet, test-control-plane  Back-off restarting failed container
```

```
kubectl describe pod prometheus-operator-5566546995-g8tsc
  Normal   Pulled     16m                  kubelet, test-control-plane  Container image "quay.io/coreos/prometheus-operator:v0.29.0" already present on machine
  Warning  BackOff    16m (x2 over 16m)    kubelet, test-control-plane  Back-off restarting failed container
  Warning  Unhealthy  98s (x448 over 16m)  kubelet, test-control-plane  Readiness probe failed: HTTP probe failed with statuscode: 503
```

morvencao:
CircleCI workflow is working now, but the steps are fixed. 
So what's you local test step? Seems that you should run the `run-prometheus-operator-config-test` target.

ericvn:
You also need https://github.com/istio/installer/pull/137 before `make test` will run to fix this.

ericvn:
Note that the prometheus operator test was never added in the CI and the CI does not do a `make test`. Once PR 137 is merged, the CI can be updated to do one or both.

ericvn:
Well, #137 should have fixed this but the prometheus operator has a half release which is causing a new failure: https://github.com/coreos/prometheus-operator/issues/2598. Seems they have a tentative plan to get that fully published on Monday.

ericvn:
The prometheus-operator should now be fixed. Running a `make test` to confirm things work, or maybe more likely, what is broken next ;) .

ericvn:
`make test` now passes so closing this issue. 

We still need to get the prometheus-operator added into the CI pipeline.

PR #137, opened 2019-05-08
fix: ensure prom operator is not installed in an auto-inject namespace

douglas-reid:
When running a full `make test`, all namespaces are selected for auto-injection, which causes issues with the prom operator install. This PR moves the prom operator to its own namespace and disables auto-injection in that namespace.

### DIFF
--- a/test/install.mk
+++ b/test/install.mk
@@ -129,10 +129,13 @@ install-policy:
 	kubectl wait deployments istio-policy -n ${ISTIO_NS} --for=condition=available --timeout=${WAIT_TIMEOUT}
 
 # This target should only be used in situations in which the prom operator has not already been installed in a cluster.
+install-prometheus-operator: PROM_OP_NS="prometheus-operator"
 install-prometheus-operator:
-	# installs the prom operator in the default namespace
-	kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/bundle.yaml
-	kubectl wait --for=condition=available --timeout=${WAIT_TIMEOUT} deploy/prometheus-operator
+	kubectl create ns ${PROM_OP_NS} || true
+	kubectl label ns ${PROM_OP_NS} istio-injection=disabled --overwrite
+	curl -s https://raw.githubusercontent.com/coreos/prometheus-operator/master/bundle.yaml | sed "s/namespace: default/namespace: ${PROM_OP_NS}/g" | kubectl apply -n ${PROM_OP_NS} -f -
+	# kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/bundle.yaml
+	kubectl -n ${PROM_OP_NS} wait --for=condition=available --timeout=${WAIT_TIMEOUT} deploy/prometheus-operator
 
 # This target expects that the prometheus operator (and its CRDs have already been installed).
 # It is provided as a way to install Istio prometheus operator config in isolation.
--- a/test/install.mk
+++ b/test/install.mk
@@ -133,9 +133,10 @@ install-prometheus-operator: PROM_OP_NS="prometheus-operator"
 install-prometheus-operator:
 	kubectl create ns ${PROM_OP_NS} || true
 	kubectl label ns ${PROM_OP_NS} istio-injection=disabled --overwrite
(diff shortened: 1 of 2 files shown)


douglas-reid:
Re-running `make test` to make sure we've solved the issue. Running the make target directly (not via KIND, etc.) is working.

costinm:
Looks like kind is broken - we need to add it to the image to get the tests passing.

test/install.mk
```diff
@@ -129,10 +129,15 @@ install-policy:
 	kubectl wait deployments istio-policy -n ${ISTIO_NS} --for=condition=available --timeout=${WAIT_TIMEOUT}
 
 # This target should only be used in situations in which the prom operator has not already been installed in a cluster.
+install-prometheus-operator: PROM_OP_NS="prometheus-operator"
```
costinm: Not sure it's common to name the namespace with -operator - maybe 'prometheus' is enough ?

test/install.mk
```diff
@@ -129,10 +129,15 @@ install-policy:
...
 install-prometheus-operator:
-	# installs the prom operator in the default namespace
-	kubectl apply -f https://raw.githubusercontent.com/coreos/prometheus-operator/master/bundle.yaml
-	kubectl wait --for=condition=available --timeout=${WAIT_TIMEOUT} deploy/prometheus-operator
+	kubectl create ns ${PROM_OP_NS} || true
```
costinm: Do you want to add this to the install.sh script, and maybe make it the default ?

test/install.mk
```diff
@@ -129,10 +129,15 @@ install-policy:
...
+	kubectl label ns ${PROM_OP_NS} istio-injection=disabled --overwrite
+	curl -s https://raw.githubusercontent.com/coreos/prometheus-operator/master/bundle.yaml | sed "s/namespace: default/namespace: ${PROM_OP_NS}/g" | kubectl apply -f -
```
costinm: Can you check in the current bundle.yaml, and maybe create a kustomize.yaml file as well ? This will allow
kubectl apply -k github.com/istio/installer/kustomize/prometeus

douglas-reid: This was an explicit choice to avoid owning any sort of install of the prometheus operator itself. Do you believe that we should own that install?

Here it is used exclusively for testing purposes.

test/install.mk
douglas-reid: The default (from the packaged bundle) is to install to "default". That seems counter to the philosophy here. I have no strong feelings on the namespace.

Does the non-operator bit install to just `prometheus`?  Should we distinguish between the two approaches?

test/install.mk
douglas-reid: If you don't mind me asking, what is the distinction between `install.sh` and the make files?

test/install.mk
ericvn: I can look after my next test pass, but the non-operator installs into `istio-control` iirc.

ericvn:
Tested this twice using `make test` and things now pass.

costinm:
Non operator should install in istio-telemetry - the tests didn't consider
this, and we temporarily changed the tests to demo-mode ( i.e. single
namespace).

If operator works - would be very happy to have it as default or single
option.

test/install.mk
howardjohn: I think this is a good point. AFAIK originally `install.sh` was intended as a simple way to install for demo usages. Now we have the `kubectl -k ...` command for that, so do we need the script?

ericvn:
fixes #135

douglas-reid:
@costinm agreed for default / single mode. more testing will be needed before that step.

Is everyone fine with updating PROM_OPT_NS to `prometheus`?

Is that the last blocker?

test/install.mk
costinm: Interesting point... I think we can have a copy of the installer. We can do 'curl -s..' part of the build process to update the copy. 

But we still want to be able to build a tar.gz or docker image with everything that is needed to run Istio - self-contained and hermetic. Our own operator will use it as well.

Think of it as 'vendoring' :-)

And we can add our own Kustomization on top of it.

howardjohn:
Mergify seems to be stuck queueing, going to merge this manually

douglas-reid:
Thanks all for the good reviews and help!

PR #139, opened 2019-05-09
fix kind issue in CI.

morvencao:
`kind` in master is broken, see: https://github.com/kubernetes-sigs/kind/issues/509, use released version as a workaround.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -278,7 +278,7 @@ git.dep: ${GOPATH}/src/istio.io/istio ${GOPATH}/src/istio.io/tools
 ${GOPATH}/bin/kind:
 	echo ${GOPATH}
 	mkdir -p ${TMPDIR}
-	go get -u sigs.k8s.io/kind
+	git clone https://github.com/kubernetes-sigs/kind ${GOPATH}/src/sigs.k8s.io/kind && cd ${GOPATH}/src/sigs.k8s.io/kind && git checkout v0.2.1 && go install .
 
 ${GOPATH}/bin/dep:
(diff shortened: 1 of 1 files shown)


costinm:
For the record - I'm not sure 0.2.1 is best option, it's from Mar27 - and may not work with the image we built.

Maybe we can checkout on SHA, from 2-3 days ago ? 

We'll also need a similar fix in the Docker image, but less urgent.

And thanks for the quick fix !

morvencao:
Yes, it's really annoying that `kind` master is broken. Let's see if `kind` v0.2.1 is stable or not.

Makefile
```diff
@@ -278,7 +278,7 @@ git.dep: ${GOPATH}/src/istio.io/istio ${GOPATH}/src/istio.io/tools
 ${GOPATH}/bin/kind:
 	echo ${GOPATH}
 	mkdir -p ${TMPDIR}
-	go get -u sigs.k8s.io/kind
+	git clone https://github.com/kubernetes-sigs/kind ${GOPATH}/src/sigs.k8s.io/kind && cd ${GOPATH}/src/sigs.k8s.io/kind && git checkout v0.2.1 && go install .
```
BenTheElder: another option if docker is available and HEAD is still desired is: 
```suggestion
	git clone https://github.com/kubernetes-sigs/kind && cd kind && make install
```
I do think we should instead get another kind release out and would request sticking to released versions for CI given the alpha state.

morvencao: Thanks, let me have a try.

PR #140, opened 2019-05-09
[WIP]fix kind build issue.

morvencao:
Follow up PR for https://github.com/istio/installer/pull/139, as it's just a short-term fix and `kind` released version `v0.2.1` might not be outdated. We would prefer to the edge kind.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -278,7 +278,7 @@ git.dep: ${GOPATH}/src/istio.io/istio ${GOPATH}/src/istio.io/tools
 ${GOPATH}/bin/kind:
 	echo ${GOPATH}
 	mkdir -p ${TMPDIR}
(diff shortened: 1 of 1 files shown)


morvencao:
@costinm The doc of `kind` has been updated with go.module, see: https://github.com/kubernetes-sigs/kind#installation-and-usage

Now they switch to the following command to get `kind`:
```
GO111MODULE="on" go get -u sigs.k8s.io/kind@master
```

However, the golang version of our circleCI is `go1.11`, which doesn't support `go get` outside of existing module.

And after https://github.com/kubernetes-sigs/kind/pull/510 is merged, I have done some local test, to build the new `kind` with go.module anywhere, we need to upgrade the go version `1.12` or greater.

howardjohn:
I think we got some fixes merge for this, or this still relevant?

costinm:
Once again - the real fix is to add the same lines to istio/istio/.circleci/Dockerfile - so we don't have to do that on each test.

costinm:
https://github.com/istio/istio/pull/13963/files

morvencao:
Fixed in https://github.com/istio/installer/pull/145, closing this.

PR #142, opened 2019-05-09
make pdb immutable.

morvencao:
IIRC, we have a limitation about the `PodDisruptionBudget `, see: https://github.com/istio/istio/issues/11210

@linsun @sdake We still can need the immutable pdb with `kubectl apply...`, right?


howardjohn:
lgtm, will let @linsun or @sdake approve though since they dealt with this in istio/istio

PR #143, opened 2019-05-09
Echo-test for minimal installation in CI.

jsPlachetka:
Adds a new job to circleci, which does the following:
- installs minimal istio
- runs the echo integration test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -284,7 +284,7 @@ jobs:
       - store_test_results:
           path: /go/out/tests
 
-  run-minimal-test:
+  run-test-on-base-installation:
     <<: *integrationRemote
 
     # Can run multiple images, first is primary, all use localhost
@@ -329,7 +329,7 @@ workflows:
       - install-minikube
       - install-demo
       - noauth
-      - run-minimal-test
+      - run-test-on-base-installation
 
   all:
     jobs:
@@ -340,4 +340,4 @@ workflows:
       - install-minikube
       - install-demo
       - noauth
-      - run-minimal-test
+      - run-test-on-base-installation
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -284,6 +284,30 @@ jobs:
       - store_test_results:
           path: /go/out/tests
 
+  run-minimal-test:
+    <<: *integrationRemote
+
+    # Can run multiple images, first is primary, all use localhost
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run: make gitdep
+      - run:
+          # Create KIND cluster (separate so we get time info)
+          command: make info dep prepare
+      - run:
+          command: make sync
+      - run:
+          command: make docker-run-test TEST_TARGET="run-minimal-test"
+      - run:
+          when: always
+          command: make kind-logs
+      - store_test_results:
+          path: /go/out/tests
+      - store_artifacts:
+          path: /go/out/logs
+
 workflows:
   version: 2
   nightly:
@@ -303,6 +327,7 @@ workflows:
       - install-minikube
       - install-demo
       - noauth
+      - run-minimal-test
 
   all:
     jobs:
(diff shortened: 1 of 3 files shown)


jsPlachetka:
I give my consent.

phil9909:
I give my consent.

howardjohn:
Its not clear to me why we want this. We already have a test that runs the echo test (and others). Is the difference here it is with the minimal install?

.circleci/config.yml
```diff
@@ -234,6 +234,30 @@ jobs:
       - store_test_results:
           path: /go/out/tests
 
+  run-minimal-test:
```
costinm: Any chance we can instead add "run-minimal-test" step to all other tests ? Just after install.

.circleci/config.yml
```diff
@@ -234,6 +234,30 @@ jobs:
...
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run: make gitdep
+      - run:
+          # Create KIND cluster (separate so we get time info)
+          command: make info dep prepare
+      - run:
+          command: make sync
+      - run:
```
costinm: Generally we split the install of a profile from running the test - mainly to make it easier to see the time and
where the failure happens.

test/install.mk
```diff
@@ -5,6 +5,10 @@
 
 INSTALL_OPTS="--set global.istioNamespace=${ISTIO_NS} --set global.configNamespace=${ISTIO_NS} --set global.telemetryNamespace=${ISTIO_NS} --set global.policyNamespace=${ISTIO_NS}"
 
+ifneq (${ENABLE_NAMESPACES_BY_DEFAULT}, false)
```
costinm: I think that is 
ENABLE_NAMESPACES_BY_DEFAULT ?= true

costinm:
Minimal test is useful because it runs a single small sanity test - without a lot of deps. So we can fail faster and save the time to run the entire suite, and it should work on all environments.

We need to find the proper test to run - make sure it works in the minimal profile - but I think  the idea is good.

.circleci/config.yml
phil9909: Would you suggest:
- installing a minimal version first, then running the minimal tests followed by the installation of the remaining components and finally running additional tests.
or
- doing a complete install, running the minimal test followed by additional tests.

test/install.mk
phil9909: done

.circleci/config.yml
phil9909: done: https://github.com/istio/installer/pull/143/commits/60a59f5563619c4f6d13457f452c7ca5ed42b9b3

jsPlachetka:
I signed it!

c0d1ngm0nk3y:
> Its not clear to me why we want this. We already have a test that runs the echo test (and others). Is the difference here it is with the minimal install?

Exactly. The naming is a bit confusing, we will try to improve. The idea is **NOT** to have a smoke test to fail fast, but to have some test with the minimal (base) installation.

c0d1ngm0nk3y:
> Minimal test is useful because it runs a single small sanity test - without a lot of deps. So we can fail faster and save the time to run the entire suite, and it should work on all environments.

@costinm Actually, the idea here is  a different one. Like @howardjohn assumed, this should verify the minimal (base) installation with some test.

c0d1ngm0nk3y:
I give my consent.
I signed it

c0d1ngm0nk3y:
How to make the cla/google bot happy? It doesn't really like pair programming, does it?

@howardjohn Any ideas what we could do better regarding the cla bot? We all have signed the cla.

.circleci/config.yml
costinm: The first - after we install the minimal test we can add another environment and run its test ( and maybe repeat the first tests as well - to make sure they don't interfere with each other ).

It would reproduce the case of a user adding a minimal, then installing other components.

I wouldn't mind if the first test leaves a fortio client/server running, and after installing more components and doing more tests checks with a final test that there are no 50x.  We are supposed to do this in the stability cluster, but it would be nice to do it as part of the install test
presubmit as well.

PR #144, opened 2019-05-09
Update kind in images to latest edge - requires Go 1.12.5 and modules

ericvn:
Kind was updated to use Go modules which requires Go 1.12.5 and modules turned on.

We will keep using the `edge` version.


ericvn:
I do have the similar changes if we wanted to use the `stable` kind release. In that case we get the binaries in the final stage and remove the _go get_ in the Golang stage.

ericvn:
Images build fine locally. Running a `make test` (with the yet un-merged prometheus operator changes) to verify the `istionightly/kind:v1.14.1-1` still works.

ericvn:
`make test` passed, removing WIP.

Issue #141, opened 2019-05-09
Upgrade golang version in CI env


morvencao:
Context:

The doc of `kind` has been updated with go.module, see: https://github.com/kubernetes-sigs/kind#installation-and-usage

Now they switch to the following command to get `kind`:
```
GO111MODULE="on" go get -u sigs.k8s.io/kind@master
```

However, the golang version of our circleCI is `go1.11`, which doesn't support `go get` very well outside of existing module.

And after https://github.com/kubernetes-sigs/kind/pull/510 is merged, I have done some local tests, to build the new `kind` with go.module anywhere, we need to upgrade the go version `1.12` or greater.

We nedd to upgrade the golang version in CI env to 1.12 or beyond.

/cc @costinm

ericvn:
I believe this can now be closed.

PR #145, opened 2019-05-09
Update CI to use Golang 1.12 and `edge` kind

ericvn:


ericvn:
This updates to  the same image Istio/Istio uses in circle in https://github.com/istio/istio/pull/13531

ericvn:
Fixes: #141

PR #148, opened 2019-05-10
import certmanager chart.

morvencao:
Part of https://github.com/istio/installer/issues/113


security/certmanager/templates/_helpers.tpl
```diff
@@ -0,0 +1,32 @@
+{{/* vim: set filetype=mustache: */}}
```
howardjohn: Doesn't look like this file is used anywhere

PR #149, opened 2019-05-10
exclude tracing test.

morvencao:
Tracing test has been added in main repo: https://github.com/istio/istio/pull/13790, the new installer hasn't imported tracing component, we need to temporarily exclude that.


morvencao:
@howardjohn Thanks.

PR #159, opened 2019-05-12
Add label to telemetry service

howardjohn:
This is needed for prometheus to scrape it properly.


howardjohn:
@phil9909 we don't use prow here so /lgtm and others don't work, just the github ui

PR #160, opened 2019-05-13
Use daily master for installer pipeline instead of 1.1

c0d1ngm0nk3y:
The install of 1.1 is not supported, so I have adjusted the pipeline to

* install daily-master
* test
* install old (fixed) version of master #until 1.2 was released
* test
* update to daily master
* test

### DIFF
--- a/.concourse/concourse.yaml
+++ b/.concourse/concourse.yaml
@@ -12,12 +12,13 @@ resources:
     type: git
     source:
       uri: "https://github.com/istio/istio-installer"
+
   - name: istio-master
     type: git
     source:
       uri: "https://github.com/istio/istio"
       branch: master
-      
+
 jobs:
   - name: install
     serial_groups: [ cluster ]
@@ -45,9 +46,9 @@ jobs:
             git remote -v
             git branch
 
-    - task: install-master-daily
-      params: &params-master-daily
-        TAG: master-latest-daily
+    - task: install-master
+      params:
+        TAG: master-daily
         HUB: gcr.io/istio-release
         KUBECONFIG: ((KUBECONFIG))
         # Remove overly generous resource request to avoid unnecessarily scaling up k8s nodes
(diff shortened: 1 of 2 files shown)


.concourse/concourse.yaml
```diff
@@ -103,22 +103,22 @@ jobs:
...
 
-  - name: install-1.1.0
+  - name: install-old-master # we probably want 1.1 here, but we would need an old installer
     serial_groups: [ cluster ]
     plan:
     - get: 24h
       trigger: true
-      passed: [test-after-install]
+      passed: [ test-after-install ]
     - get: istio-installer
```
jkbschmid: If we use istio-installer master to install a fixed, older istio version from which we can do an upgrade, I suggest using a fixed older version of the installer as well. Otherwise we have to ensure that the installer master is still compatible with this fixed, older version.

c0d1ngm0nk3y:
@holgero I will try an older version of the installer to install 1.1. Probably before b7e869e69f8755af913ae19763e7483b01db0974. But there is no supported way to have 1.1 installed in several namespaces, right?

c0d1ngm0nk3y:
This state now installs istio-1.1 and it works, but it cannot update to master anymore:

```
The Deployment "grafana" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"release":"istio-telemetry-istio-grafana", "app":"grafana"}: `selector` does not match template `labels`
```

Is this supported/should work at all?

.concourse/concourse.yaml
```diff
@@ -132,22 +132,28 @@ jobs:
           args:
           - -ec
           - |
+            # see github issue #65, so we take an older version that works with 1.1
+            cd istio-installer
+            git fetch origin master
+            git checkout 02d6d54442fb1f48528742a40a37c3a37888db79
```
howardjohn: HEAD should (and absolutely, P0) needs to work with release-1.1. 

What happens when you use the latest, we should be able to install alongside an existing version with no interference?

howardjohn:
@c0d1ngm0nk3y being able to upgrade is the main purpose of the installer, so that is definitely an issue. 

Is that error when installing master when 1.1 is running? Is it installed in istio-telemetry?

costinm:
+1 on what @howardjohn  said. 

We need istio-1.1 in istio-system, and new installer in istio-control+istio-telemetry, etc.

.concourse/concourse.yaml
jkbschmid: > HEAD should (and absolutely, P0) needs to work with release-1.1.



Just to clarify: You mean that **upgrading** from an installation of release-1.1 should work with the current installer, correct? Installing 1.1 with the new installer is not supported according to [this issue](https://github.com/istio/istio/issues/14146).

> What happens when you use the latest, we should be able to install alongside an existing version with no interference?

 
When we use the latest version of the new installer to upgrade from 1.1, the error that @c0d1ngm0nk3y mentioned occurs. To install alongside an existing version, we would have to modify this pipeline to use a canary-style upgrade (and not in-place) for all other components as well (so far, only istio-control is upgraded like this, i.e. by installing separately + switching).

jkbschmid:
> Is that error when installing master when 1.1 is running?

@howardjohn Yes.

> Is it installed in istio-telemetry?

Yes, in the pipeline the new installer installs/upgrades in istio-telemetry. Is this not as intended? If not, how is switching between the old and the new version supposed to work and where is this going to be implemented (istio-ctl?) ?

As a side note: Is there a Readme/design doc that I can have a look at which outlines the purpose of all these different tools during installation, testing and upgrade (make/helm/istioctl/kustomize/operator)? We would like to avoid a pipeline that tests a method of installation (install script/iop) that is not actually used (see [this issue](https://github.com/istio/istio/issues/14117)).

costinm:
To clarify a bit: 
- using helm/tiller for install - should work, but user needs to run multiple helm commands, one per a-la-carte component.

- using kustomize/kubectl -k - should work for the demo and default profile, possibly others. This is what k8s SIG recommends.

- the iop script can be used to install component by component for full 'a-la-carte' 

- Makefile is meant for testing the repo, not to be used by users. Should have targets using iop to generate what is needed in kustomize.

- install.sh - my understanding is that it is meant for concourse and other tests. I guess it could be 
used as a wrapper around installing everything and alternative to 'apply -k' - but I'm not sure why
is better than either iop component by component or the 'pre-built' profiles. 

- operator - will use the templates from installer, but have more automation. That's what will do 
the equivalent of 'install.sh' when ready, i.e. add/remove all needed components. 

- istioctl - the goal is to fold 'iop' into istioctl.   

That's my understanding at least, from the many design docs and meetings.

costinm:
I'm not sure I understand the change - but the description is certainly not what we want to do.
We must test:
- install release-1.1 ( with old installer, in istio-system )
- install app1 - verify it works
- install master - in different namespaces
- install an app2 pointing to master (istio-control), verify it works
- verify app1 still works( ideally use fortio and check no errors )
- migrate app1 to istio-control via rolling upgrade, verify it works and no errors.

c0d1ngm0nk3y:
> * install master - in different namespaces

@costinm Big thanks for this clarification. 3 follow-up questions:
1) Since `citadel` is in both cases installed in `istio-system`, this works only because the citadel service has different names, right? By design?
2) How does this work with CRDs? They are not namespaced, right?
3) What if some istio configuration objects (e.g. `virtual service`) refer to for example `istio-ingress.istio-system`? Can you adapt the configuration first and then switch the istio version?

holgero:
> @c0d1ngm0nk3y being able to upgrade is the main purpose of the installer, so that is definitely an issue.

So I think we should look into this and fix it.

c0d1ngm0nk3y:
`kustomize` can only be used for the demo scenarios, right? We want to test the "a-la-carte" variant here.  The `install.sh` basically wraps the `iop` commands.

Since there seems to be no update case, we removed the update steps for now. The pipeline checks now if the master of installer can be used with daily-master of  istio. Makes sense?

It is triggered
* for each merged commit (PR validation comes next)
* every 24h (because of daily-master)

c0d1ngm0nk3y:
Again the &lt;sarcasm> very helpful </sarcasm&gt; cla bot :( What is the issue? Can anyone say what should we do differently? Pair programming does not really work with this bot, does it?

linsun:
@c0d1ngm0nk3y  have you signed the CLA or verify consent?

c0d1ngm0nk3y:
> @c0d1ngm0nk3y have you signed the CLA or verify consent?

@linsun From what I understand, the bot has noticed that everyone involved signed the CLA (which is right). I think my consent is clear since I started this PR, so the other two need to add a comment? But how should pair programming work with this bot and why *does* it sometimes work and sometimes, the bot blocks the PR? I would love to understand the rules, so we might optimise our working model.

@jkbschmid @jsPlachetka Could you comment your consent?

jkbschmid:
I consent.

jsPlachetka:
I give my consent

Issue #49, opened 2019-04-23
README.md expects $IBASE to be set but doesn't explain how to set it


sdake:
`$IBASE` should probably be set to `$GOPATH/src/istio-ecosystem/istio-installer` however this is not documented AFIAK.

Cheers
-steve

phil9909:
I was about to change the README for #49. Previously, the commands in the README worked after sourcing `env.sh`. This is no longer true (unfortunately `env.sh` does no longer set `$IBASE`).
I see multiple options here:
1. Document how to set `$IBASE` manually.
1. Update `env.sh` to set `$IBASE` again and update README with instructions to source `env.sh`
1. Get rid of `$IBASE`: 
   - `bin/iop` could search for helm-charts relative to it's own location (using `${BASH_SOURCE[0]}`)
   - of course this would require the commands in the README to be updated.

@sdake @costinm any opinions/preferences?

PR #164, opened 2019-05-13
Allow bin/iop to be sourced without exiting bash

phil9909:
Our README is full of commands like this `iop istio-control istio-autoinject $IBASE/istio-control/istio-autoinject --set enableNamespacesByDefault=true` which in my understanding require bin/iop to be sourced first. This worked fine in the past, but is no longer possible as iop calls `exit 1` when called without arguments.

This PR enables sourcing of iop again:
- If the script is being sourced, it will call `return` after all functions and variables (e.g. `$IBASE`) are set.
- Additionally, before calling `set -x` the current state of xtrace is stored and restored later.

### DIFF
--- a/bin/iop
+++ b/bin/iop
@@ -2,6 +2,7 @@
 
 # Example script to install - will be replaced by operator or other tools, used for testing and as sample.
 
+RESTORE_SET_X=$(shopt -po xtrace)
 set -x
 
 BASE=${BASE:-`pwd`}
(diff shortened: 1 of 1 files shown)


howardjohn:
why not just run `bin/iop` or add iop to your path? I don't think it was intended to be sourced.

I am also fine with the change though

phil9909:
You're right. [bin/env.sh](../blob/master/bin/env.sh) was the one to be sourced.
To give you some context, I was about to change the README for #49. Previously, the commands in the README worked after sourcing `env.sh`. This is no longer true (unfortunately `env.sh` does no longer set `$IBASE`).
See my comment: https://github.com/istio/installer/issues/49#issuecomment-492126434

PR #165, opened 2019-05-13
Add neverInjectSelector and alwaysInjectSelector properties to Helm chart

cagataygurturk:
Capability to specify alwaysInjectSelector and neverInjectSelector in the Helm chart.

Resolves istio/istio#14020


istio-control/istio-autoinject/templates/sidecar-injector-configmap.yaml
```diff
@@ -14,5 +14,9 @@ data:
 
   config: |-
     policy: {{ .Values.global.proxy.autoInject }}
+    alwaysInjectSelector:
+{{ toYaml .Values.sidecarInjectorWebhook.alwaysInjectSelector | indent 20}}
```
howardjohn: should be indented 6? I guess it is a list so doesn't matter much

PR #170, opened 2019-05-14
add nodeSelector and toleration.

morvencao:
- add nodeSelector
- add tolerations
- fix a few yaml format issues

For `nodeSelector` and `toleration`, define default values in global section to be applied to all deployments/daemonsets so that all pods can be scheduled to a particular nodes. Each component can overwrite these default values by adding its own block in the relevant section below and setting the desired values.

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -251,6 +251,12 @@ global:
   # the desired values.
   defaultNodeSelector: {}
 
+  # Default node tolerations to be applied to all deployments so that all pods can be 
+  # scheduled to a particular nodes. Each component can overwrite these default 
+  # values by adding its tolerations block in the relevant section below and setting 
+  # the desired values.
+  defaultTolerations: []
+
   # Whether to perform server-side validation of configuration.
   configValidation: true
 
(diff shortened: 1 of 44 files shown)


howardjohn:
The change looks good, and we definitely need some of the fixes that got added here.

My concern is that Kustomize is the perfect solution for toleration, affinity etc. We already support affinty so we should probably keep that, but do we want to add more (fairly complicated) fields that can be instead replaced by Kustomize? Given we already have affinity maybe it makes sense to get toleration as well (as in: we should have toleration+affinity or neither, not one or the other).

@costinm @sdake any thoughts?

global.yaml
```diff
@@ -251,6 +251,12 @@ global:
   # the desired values.
   defaultNodeSelector: {}
 
+  # Default node tolerations to be applied to all deployments so that all pods can be
```
costinm: I don't think this makes any sense with the new installer - each component is installed independently, and it would have different settings.  We also want to move this kind of 'global override' to kustomize and reduce API surface in values.yaml

costinm:
Looks good otherwise.

global.yaml
morvencao: remove global default value.

morvencao:
Can we get this merged? Since it also contains quite a few some bugs about template.

PR #181, opened 2019-05-16
Sync ingressgateway with master

howardjohn:
@costinm I kept in all of your changes


gateways/istio-ingress/templates/deployment.yaml
```diff
@@ -92,18 +96,23 @@ spec:
...
+        {{- end}}
+        {{- if .Values.global.proxy.componentLogLevel }}
+          - --proxyComponentLogLevel={{ .Values.global.proxy.componentLogLevel }}
+        {{- end}}
+        {{- if .Values.global.logging.level }}
+          - --log_output_level={{ .Values.global.logging.level }}
+        {{- end}}
           - --drainDuration
-          - '{{ $gateway.drainDuration }}'
+          - '45s' #drainDuration
```
costinm: Can you add back the option ? For ingress it's important to be able to customize this.
Same for connect timeout

mousavian:
It seems this PR has broke installation (`install_all`) because this commit is not available in default 1.1.4 tag.

I'm getting following error from ingressgateway:
```bash
$kubectl log istio-ingressgateway-868fc8d7fd-9n4cv -n istio-ingress -p
2019-05-28T04:52:03.543645Z	error	unknown flag: --proxyComponentLogLevel
Error: unknown flag: --proxyComponentLogLevel
```

PR #182, opened 2019-05-17
verify statefulset for prometheus-operator test 

morvencao:
See: https://github.com/coreos/prometheus-operator/blob/master/Documentation/design.md#prometheus
> For each Prometheus resource, the Operator deploys a properly configured StatefulSet in the same namespace. The Prometheus Pods are configured to mount a Secret called <prometheus-name> containing the configuration for Prometheus.

We should check the promesheus **statefulset**, instead of **deployment**, as prometheus-operator will create **statefulset** for each `Prometheus` resource, not **deployment**.


morvencao:
Update: I found it's not convenient to check the status of prometheus statefuset directly, let's switch to check the pod status since we know its name.

PR #183, opened 2019-05-17
fix(dashboards): update grafana and repair broken performance dash

douglas-reid:
This PR accomplishes the following:
- updates the default image of grafana to the latest release (6.0.2 -> 6.1.6).
- updates the performance dashboard by:
   - repairing the broken panels that had gone missing
   - converting to consistent use of rows
   - moving the README to a collapsed-by-default panel at the top
   - adding perf-relevant panels for individual components to the bottom (as inspired by @Stono).
- **converts the places where `istio-system` was hardcoded to use `istio-control` as the namespace.** This will make the perf dashboards work for traditional installs, but may cause different results when components are deployed in distinct namespaces.
- **converts the workload names from `istio-ingressgateway` to `ingressgateway`**. This may mean additional work is required to back-port to older installs of Istio. More work will be necessary here most likely.

Reviewers, please take a look an experiment with this update in your own clusters.  Snapshot: https://snapshot.raintank.io/dashboard/snapshot/P8j00RI9hGJxw0GaOzUsGb4FqH4PMkxz

### DIFF
--- a/istio-telemetry/grafana/dashboards/istio-performance-dashboard.json
+++ b/istio-telemetry/grafana/dashboards/istio-performance-dashboard.json
@@ -1,40 +1,4 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
-      "label": "Prometheus",
-      "description": "",
-      "type": "datasource",
-      "pluginId": "prometheus",
-      "pluginName": "Prometheus"
-    }
-  ],
-  "__requires": [
-    {
-      "type": "grafana",
-      "id": "grafana",
-      "name": "Grafana",
-      "version": "5.2.3"
-    },
-    {
-      "type": "panel",
-      "id": "graph",
-      "name": "Graph",
-      "version": "5.0.0"
-    },
-    {
-      "type": "datasource",
-      "id": "prometheus",
-      "name": "Prometheus",
-      "version": "5.0.0"
-    },
-    {
-      "type": "panel",
-      "id": "text",
-      "name": "Text",
-      "version": "5.0.0"
-    }
-  ],
   "annotations": {
     "list": [
       {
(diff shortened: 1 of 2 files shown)


howardjohn:
> converts the workload names from istio-ingressgateway to ingressgateway. This may mean additional work is required to back-port to older installs of Istio. More work will be necessary here most likely.

This comes from the `app` label right? If so I think it should be `istio-ingressgateway`, we had it as ingressgateway because we hadn't synced the template with master yet, should all be consistently `istio-ingressgateway` now.

> converts the places where istio-system was hardcoded to use istio-control as the namespace. This will make the perf dashboards work for traditional installs, but may cause different results when components are deployed in distinct namespaces.

I wonder if we can just use helm here and get the values? Sounds like it would be a pain, but looks like `{{` is only used for the legends so maybe wouldn't be too bad... @costinm any thoughts?


The dashboard itself looks good to me. @mandarjog any thoughts?

douglas-reid:
> This comes from the `app` label right? If so I think it should be `istio-ingressgateway`, we had it as ingressgateway because we hadn't synced the template with master yet, should all be consistently `istio-ingressgateway` now.

I'd have to double check, but I believe this is coming from the `workload` label, which I believe is coming from the deployment name. I was testing from HEAD.

howardjohn:
@douglas-reid we should change that too then, I think it was not intentionally changed? No need to change the name of something, just adds to more complications (like this). I can get out a PR to fix that

howardjohn:
https://github.com/istio/installer/pull/185 fixes the names

howardjohn:
Now we change the names already in #185 so we will need to update them in another pr I think.

I changed my mind about helm though, that would make developing the dashboards a bit of a pain I think. Not sure a better approach though...

PR #185, opened 2019-05-17
Fix gateway names

howardjohn:
These should match what we have in master and be consistent with other
components.


costinm:
You missed a 'make build' or update for the demo. Good we have a test.

costinm:
Which was not required apparently. We'll need a separate PR to fix demo

PR #189, opened 2019-05-19
Remove prometheus metric_relabel_configs

mandarjog:
We control the metrics that are emitted at the proxy.
This configuration is no longer needed.


costinm:
Rebase ?

PR #190, opened 2019-05-20
fix(dashboards): resolve a few minor issues around job names and queries

douglas-reid:
This PR adopts the work in https://github.com/istio/istio/pull/10934 and corrects some of Mixer Pilot confusion introduced in earlier PRs.

### DIFF
--- a/istio-telemetry/grafana/dashboards/istio-performance-dashboard.json
+++ b/istio-telemetry/grafana/dashboards/istio-performance-dashboard.json
@@ -1373,7 +1373,7 @@
       "steppedLine": false,
       "targets": [
         {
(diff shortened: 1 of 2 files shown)


istio-telemetry/grafana/dashboards/istio-performance-dashboard.json
```diff
@@ -1373,7 +1373,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "process_virtual_memory_bytes{job=\"pilot\"}",
```
howardjohn: I think I missed this on istio/istio, good catch. Hard to tell these issues from a huge json file

PR #195, opened 2019-05-22
only add necessary labels for istio-cni.

morvencao:
Labels are heavy for k8s resources, let's only add necessary labels for istio-cni to avoid upgrade issues.


istio-cni/templates/daemonset.yaml
```diff
@@ -26,7 +26,8 @@ spec:
         # marks the pod as a critical add-on, ensuring it gets
         # priority scheduling and that its resources are reserved
         # if it ever gets evicted.
-        scheduler.alpha.kubernetes.io/critical-pod: ''
+        scheduler.alpha.kubernetes.io/critical-pod: ""
+        sidecar.istio.io/inject: "false"
```
howardjohn: did we want to add this? it seems right but not sure why we don't have it before? just because it was in Istio system which was ignored by injector?

istio-cni/values_gke.yaml
```diff
@@ -1,18 +1,19 @@
-hub: docker.io/tiswanso
-tag: v0.1-dev
-pullPolicy: Always
+cni:
+  hub: docker.io/tiswanso
```
howardjohn: not related to this change but this doesn't seem right

istio-cni/templates/daemonset.yaml
morvencao: OK, Yes, the annotation is only used by `isio-sidecar-injector`, in `CNI` mode, we don't need that.

istio-cni/values_gke.yaml
morvencao: Should i need to address this in separate PR?

morvencao:
@howardjohn OK, got it.

istio-cni/values_gke.yaml
tiswanso: we could use docker.io/istiocni or gcr.io like:
```
cni:
  hub: gcr.io/istio-release
  tag:  master-latest-daily
```

PR #197, opened 2019-05-22
Add ability to send annotations as env vars automagically

mandarjog:
for a gateway pod-annotations must also be set in the env.


costinm:
Never mind - not cut yet.

PR #199, opened 2019-05-23
Enable auto-backport

howardjohn:


costinm:
It seems some refactoring is blocking auto-merge.

can't load package: package istio.io/istio/tests/integration/echo: cannot find package "istio.io/istio/tests/integration/echo" in any of:...

PR #201, opened 2019-05-24
fix configNamespace for telemetry

mandarjog:
telemetry is broken due to this.

### DIFF
--- a/kustomize/istio-telemetry/mixer-telemetry.yaml
+++ b/kustomize/istio-telemetry/mixer-telemetry.yaml
@@ -486,7 +486,7 @@ spec:
           - unix:///sock/mixer.socket
           - --log_output_level=default:info
           - --configStoreURL=mcp://localhost:15019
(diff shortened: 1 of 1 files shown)


kustomize/istio-telemetry/mixer-telemetry.yaml
```diff
@@ -486,7 +486,7 @@ spec:
           - unix:///sock/mixer.socket
           - --log_output_level=default:info
           - --configStoreURL=mcp://localhost:15019
-          - --configDefaultNamespace=istio-control
```
howardjohn: I thought this was the namespace of galley? is this where the configs like CRDs are stored?

mandarjog: This is the "root" namespace as seen by mixer. for telemetry related stuff, we do not need it to be istio-control.

howardjohn:
... which is apparently not required test

PR #203, opened 2019-05-25
Add cadvisor service monitor, http only

mandarjog:
resolves https://github.com/istio/istio/issues/14164


eskp:
Wouldn't this clash with the kubelet jobs that come with Prometheus Operator?

PR #207, opened 2019-05-26
Add templated stack driver config

mandarjog:
Note: This sd config is templatized and made multi namespace friendly.


istio-telemetry/mixer-telemetry/templates/stack-driver.yaml
```diff
@@ -0,0 +1,820 @@
+{{- if .Values.mixer.adapters.stackdriver.enabled }}
```
bianpengyuan: Could you rename the file to stackdriver.yaml?

mandarjog: done.

istio-telemetry/mixer-telemetry/templates/stackdriver.yaml
```diff
@@ -0,0 +1,820 @@
...
+  compiledAdapter: stackdriver
+  params:
+    pushInterval: 10s
+    metricInfo:
+      server-request-count.instance.{{ .Release.Namespace }}:
+        # Due to a bug in gogoproto deserialization, Enums in maps must be
+        # specified by their integer value, not variant name. See
+        # https://github.com/googleapis/googleapis/blob/master/google/api/metric.proto
+        # MetricKind and ValueType for the values to provide.
+        kind: 3 # CUMULATIVE
```
cagataygurturk: @mandarjog 

[istio.io/service/server/request_count](https://cloud.google.com/monitoring/api/metrics_istio) is DELTA but it is defined as CUMULATIVE here. istio/istio#22373 seems to be resolved when this is set to DELTA (2). Can you confirm?

cagataygurturk: Sorry, it seems after setting to DELTA, mixer gives the following error:

```
One or more TimeSeries could not be written: Metric kind for metric istio.io/service/server/request_count must be CUMULATIVE, but is DELTA
````

however Google's documentation says istio.io/service/server/request_count is DELTA.

PR #208, opened 2019-05-27
Fix run-prometheus-operator-config-test.

phil9909:
See https://circleci.com/gh/istio/installer/3788 and https://circleci.com/gh/istio/installer/3787
Broken since https://github.com/istio/installer/commit/366723f122f9cb0db8c49ac01d26d31991abb0d0


c0d1ngm0nk3y:
Should be quite uncontroversial, right? I wonder why it failed that long without someone noticing...

PR #209, opened 2019-05-28
fix job names to fix performance and other dashboards

mandarjog:


mandarjog:
Perhaps the legend format does not need to change, however it is good to be consistent.
All the job / pod names have an istio- prefix now.
So without these changes the dash board is broken.

howardjohn:
> Perhaps the legend format does not need to change, however it is good to be consistent.
> All the job / pod names have an istio- prefix now.
> So without these changes the dash board is broken.

What I mean is, we shouldn't be changing the pod names. If we did accidentally change the pod names, then we should change them to match istio/istio, rather than updating the dashboards

costinm:
I'm not sure which is the best option  - given the fact the old->new
installer migration will use new namespaces, we have
a chance to fix some of the naming and make it consistent. Since TOC
decided this will be alpha in 1.2, it's even less
important to keep the 1.1 names.

After we ship in 1.3 - the names, labels, etc will likely be frozen and
hard to touch for a long time.

I would even go after the names of the containers in each pod, to make them
consistent and easier to
write CLI ( for example make sure all envoy sidecars are in containers
called istio-proxy, and the container
name is either matching the name of the deployment or uses a common name
like 'app' ).

mandarjog:
@costinm were the names changed mistakenly?
They *are* more consistent now, but it will create problems upgrading from old deployment to new .

howardjohn:
I am fine with changing the labels as long as we go through them all and explicitly set the to what we want them to be. But we shouldn't accidentally change a few of them

costinm:
Re. upgrade problems - that's the point, the old single-namespace to multi-namespace upgrade allows changing labels and job names without breakage.

After we switch to multi-namespace - we'll be again restricted on what changes we can make. 

Let's get this merged, since 'consistency for names and labels across istio components' is a very different issue. But please consider making a separate PR to improve the naming consistency since
we don't know when the next change will be - and it hurts usability.

costinm:
Sorry, added do not merge since there seem to be an istio-istio-ingressgateway. Please fix.

istio-telemetry/grafana/dashboards/galley-dashboard.json
```diff
@@ -296,14 +296,14 @@
...
-          "expr": "sum(rate(container_cpu_usage_seconds_total{container_name=~\"galley\", pod_name=~\"istio-galley-.*\"}[1m]))",
+          "expr": "sum(rate(container_cpu_usage_seconds_total{job=\"kubernetes-cadvisor\",container_name=~\"galley\", pod_name=~\"istio-galley-.*\"}[1m]))",
           "format": "time_series",
           "intervalFactor": 2,
           "legendFormat": "Total (k8s)",
           "refId": "A"
         },
         {
-          "expr": "sum(rate(container_cpu_usage_seconds_total{container_name=~\"galley\", pod_name=~\"istio-galley-.*\"}[1m])) by (container_name)",
+          "expr": "sum(rate(container_cpu_usage_seconds_total{job=\"kubernetes-cadvisor\",container_name=~\"galley\", pod_name=~\"istio-galley-.*\"}[1m])) by (container_name)",
```
howardjohn: do we need this in istio/istio as well? I was seeing the same duplicate problem on old installer. maybe affects 1.1 as well if it is really just a Kubernetes 1.13+ issue

istio-telemetry/grafana/dashboards/istio-performance-dashboard.json
```diff
@@ -323,17 +323,17 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "(sum(container_memory_usage_bytes{pod_name=~\"istio-telemetry-.*\"}) / (sum(irate(istio_requests_total[1m])) / 1000)) / (sum(irate(istio_requests_total{source_workload=\"istio-ingressgateway\"}[1m])) >bool 10)",
+          "expr": "(sum(container_memory_usage_bytes{pod_name=~\"istio-telemetry-.*\"}) / (sum(irate(istio_requests_total[1m])) / 1000)) / (sum(irate(istio_requests_total{source_workload=\"istio-istio-ingressgateway\"}[1m])) >bool 10)",
```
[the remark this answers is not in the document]
mandarjog: Good catch. Updated.

istio-telemetry/grafana/dashboards/galley-dashboard.json
mandarjog: yes, this is needed in istio/istio as well.

PR #210, opened 2019-05-28
Make number of certmanager replicas configurable

Lysholm:
This fixes: https://github.com/istio/istio/issues/14380

### DIFF
--- a/security/certmanager/templates/deployment.yaml
+++ b/security/certmanager/templates/deployment.yaml
@@ -7,7 +7,7 @@ metadata:
     app: certmanager
     release: {{ .Release.Name }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.certmanager.replicaCount }}
   selector:
     matchLabels:
       app: certmanager
--- a/security/certmanager/values.yaml
+++ b/security/certmanager/values.yaml
@@ -5,6 +5,7 @@ certmanager:
   # gateway must be updated by adding 'secretVolumes'. After the gateway
   # restart, DestinationRules can be created using the ACME-signed certificates.
   enabled: false
+  replicaCount: 1
   hub: quay.io/jetstack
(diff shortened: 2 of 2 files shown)


wattli:
CertManager does not interact with Citadel directly. As long as the multiple instances do not have conflict when generating secrets, then it should be good.

Lysholm:
@howardjohn: certmanager has had leader election since 2017: https://github.com/jetstack/cert-manager/pull/42.

I've been scaling it up since i started using it about a year ago. The non leader only has this in it's logs:
```
I0528 09:54:46.234790       1 start.go:81] starting cert-manager v0.6.2 (revision f5e1477bd7ced69e53a233484905fea16bf4102f)
I0528 09:54:46.255302       1 controller.go:141] Using the following nameservers for DNS01 checks: [10.0.0.10:53]
I0528 09:54:46.256696       1 leaderelection.go:193] attempting to acquire leader lease  istio-system/cert-manager-controller...
```
So i think it's safe but i haven't tested it extensively.

wattli:
In general i think it should be fine, but keep in mind, leaderelection does not guarantee conflict free https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L19

Lysholm:
> In general i think it should be fine, but keep in mind, leaderelection does not guarantee conflict free https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L19

I was not aware of that, thanks!

linsun:
@morvencao sorry I didn't see this earlier, looks like cert mgr ok for replica 1+.  We probably want to revert your PR on PDB for it, WDYT?

morvencao:
@linsun I did see this PR, either.  I'm OK to add back PDB for cert-manager if its replicas is configurable.

PR #212, opened 2019-05-28
Proposal to increase default replica counts when default PDB is in use

Lysholm:
This is a proposal for "fixing": https://github.com/istio/istio/issues/12602

I based the logic on what i found in `gateways/istio-egress`.

This is just a proposal, another option could be to set `defaultPodDisruptionBudget.enabled = false` in `global.yaml` or update documentation to the effect that default installs must be scaled before node maintenance is possible.

### DIFF
--- a/gateways/istio-egress/templates/autoscale.yaml
+++ b/gateways/istio-egress/templates/autoscale.yaml
@@ -10,7 +10,15 @@ metadata:
     release: {{ .Release.Name }}
 spec:
   maxReplicas: {{ $gateway.autoscaleMax }}
+{{- if $gateway.autoscaleMin }}
   minReplicas: {{ $gateway.autoscaleMin }}
+{{- else }}
+{{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+  minReplicas: 2
+{{- else }}
+  minReplicas: 1
+{{- end }}
+{{- end }}
   scaleTargetRef:
     apiVersion: apps/v1
     kind: Deployment
--- a/gateways/istio-ingress/values.yaml
+++ b/gateways/istio-ingress/values.yaml
@@ -38,9 +38,9 @@ gateways:
       name: tls
 
     # Scalability tunning
-    # replicaCount: 1
+    # replicaCount: 2
     autoscaleEnabled: true
-    autoscaleMin: 1
+    # autoscaleMin: 2
     autoscaleMax: 5
 
     cpu:
--- a/istio-control/istio-config/values.yaml
+++ b/istio-control/istio-config/values.yaml
@@ -1,6 +1,6 @@
 galley:
   image: galley
-  replicaCount: 1
+  # replicaCount: 2
 
   resources:
     requests:
--- a/gateways/istio-egress/templates/deployment.yaml
+++ b/gateways/istio-egress/templates/deployment.yaml
@@ -13,7 +13,11 @@ spec:
 {{- if $gateway.replicaCount }}
   replicas: {{ $gateway.replicaCount }}
 {{- else }}
+{{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+  replicas: 2
+{{- else }}
   replicas: 1
+{{- end }}
 {{- end }}
 {{- end }}
   selector:
--- a/gateways/istio-egress/values.yaml
+++ b/gateways/istio-egress/values.yaml
@@ -25,9 +25,9 @@ gateways:
 
 
     # Scalability tunning
-    # replicaCount: 1
+    # replicaCount: 2
     autoscaleEnabled: true
-    autoscaleMin: 1
+    # autoscaleMin: 2
     autoscaleMax: 5
     resources:
       requests:
--- a/gateways/istio-ingress/templates/autoscale.yaml
+++ b/gateways/istio-ingress/templates/autoscale.yaml
@@ -10,7 +10,15 @@ metadata:
     release: {{ .Release.Name }}
 spec:
   maxReplicas: {{ $gateway.autoscaleMax }}
+{{- if $gateway.autoscaleMin }}
   minReplicas: {{ $gateway.autoscaleMin }}
+{{- else }}
+{{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+  minReplicas: 2
+{{- else }}
+  minReplicas: 1
+{{- end }}
+{{- end }}
   scaleTargetRef:
     apiVersion: apps/v1
     kind: Deployment
--- a/gateways/istio-ingress/templates/deployment.yaml
+++ b/gateways/istio-ingress/templates/deployment.yaml
@@ -13,7 +13,11 @@ spec:
 {{- if $gateway.replicaCount }}
   replicas: {{ $gateway.replicaCount }}
 {{- else }}
+{{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+  replicas: 2
+{{- else }}
   replicas: 1
+{{- end }}
 {{- end }}
 {{- end }}
   selector:
--- a/istio-control/istio-autoinject/templates/deployment.yaml
+++ b/istio-control/istio-autoinject/templates/deployment.yaml
@@ -8,7 +8,15 @@ metadata:
     release: {{ .Release.Name }}
     istio: sidecar-injector
 spec:
+{{- if .Values.sidecarInjectorWebhook.replicaCount }}
   replicas: {{ .Values.sidecarInjectorWebhook.replicaCount }}
+{{- else }}
+{{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+  replicas: 2
+{{- else }}
+  replicas: 1
+{{- end }}
+{{- end }}
   selector:
     matchLabels:
       istio: sidecar-injector
--- a/istio-control/istio-autoinject/values.yaml
+++ b/istio-control/istio-autoinject/values.yaml
@@ -2,7 +2,7 @@ sidecarInjectorWebhook:
   # sidecar-injector webhook configuration.
   # If down, new pods will fail to start or restart for a short time, but they will
   # retry.
-  replicaCount: 1
+  # replicaCount: 2
 
   image: sidecar_injector
 
--- a/istio-control/istio-config/templates/deployment.yaml
+++ b/istio-control/istio-config/templates/deployment.yaml
@@ -8,7 +8,15 @@ metadata:
     istio: galley
     release: {{ .Release.Name }}
 spec:
+{{- if .Values.galley.replicaCount }}
   replicas: {{ .Values.galley.replicaCount }}
+{{- else }}
+{{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+  replicas: 2
+{{- else }}
+  replicas: 1
+{{- end }}
+{{- end }}
   selector:
     matchLabels:
       app: galley
--- a/istio-control/istio-discovery/templates/autoscale.yaml
+++ b/istio-control/istio-discovery/templates/autoscale.yaml
@@ -9,7 +9,15 @@ metadata:
     release: {{ .Release.Name }}
 spec:
   maxReplicas: {{ .Values.pilot.autoscaleMax }}
+{{- if .Values.pilot.autoscaleMin }}
   minReplicas: {{ .Values.pilot.autoscaleMin }}
+{{- else }}
+{{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+  minReplicas: 2
+{{- else }}
+  minReplicas: 1
+{{- end }}
+{{- end }}
   scaleTargetRef:
     apiVersion: apps/v1
     kind: Deployment
--- a/istio-control/istio-discovery/templates/deployment.yaml
+++ b/istio-control/istio-discovery/templates/deployment.yaml
@@ -18,7 +18,11 @@ spec:
 {{- if .Values.pilot.replicaCount }}
   replicas: {{ .Values.pilot.replicaCount }}
 {{- else }}
+{{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+  replicas: 2
+{{- else }}
   replicas: 1
+{{- end }}
 {{- end }}
 {{- end }}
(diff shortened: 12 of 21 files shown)


gateways/istio-ingress/values.yaml
```diff
@@ -38,9 +38,9 @@ gateways:
       name: tls
 
     # Scalability tunning
-    # replicaCount: 1
+    # replicaCount: 2
     autoscaleEnabled: true
-    autoscaleMin: 1
```
howardjohn: Doesn't seem like this needs to be commented out, or changed to 2. This affects autoscaling so it can be set to 1 and scale up to 2 fine

istio-control/istio-config/values.yaml
```diff
@@ -1,6 +1,6 @@
 galley:
   image: galley
-  replicaCount: 1
+  # replicaCount: 2
```
howardjohn: Can galley safely run with multiple replicas? @costinm

gateways/istio-ingress/values.yaml
loewenstein: I wouldn't expect the HPA and PDB logic play too well together... The former scales on load, the latter prevents scale to zero. If there is no load I'd expect HPA scales to autoscaleMin, if there is only one replica PDB prevents node drain (to keep that one instance).

loewenstein: Another question... Is the default supposed to be highly available? @costinm?

If so we'd need an autoscaleMin of 3. Otherwise we would only account for planned downtime like node maintenance, not for a failure during planning downtime.

howardjohn: Kubernetes won't scale up for the node drain? I am a bit suspicious of that, with a caveat that I have never tried it or seen docs on it, so it is mostly just a guess. 

Setting our default replicas to 2 is not a trivial change for an issue that may or may not exist for other users. At the least, can you provide some references that this is the right fix for this issue?

loewenstein: I didn't try either. My expectation towards Kubernetes is just that they haven't coupled HPA and PDB.

Lysholm: I can confirm it hadn't drained the HPA scaled pods after 2 hours on GKE 1.12.7 and GKE 1.13.6. I may have misconfigured something, but it was a pretty bog standard clean cluster created for the test, ditto istio install.

I may have had a pod without HPA on the node aswell and that prevented some scheduling logic, I will retest with only HPA.

Lysholm: @howardjohn: After scanning through the PRs and issues related to introducing the PDB I'm less sure of what's the correct solution here. They were introduced to alleviated downtime during upgrade and that's a great use case. But i think it was a mistake to enable them by default.

Disruption budgets is to me conceptually something you layer on top of HA when disruption in planned maintenance windows also need to be mitigated. As it's set up today it might give false impression to users. I can't speak for anyone else but i think it's common to test involuntary disruption by simulating it through voluntary disruptions.

IE in the original issue that lead to default disruption budgets: https://github.com/istio/istio/issues/9951
It seems this user has only one replica running and it was critical for them not to interrupt istio during maintenance. I'd wager it's more critical for them not to have it interrupted outside maintenance, but that possibility may have been obscured for them.

The other option, HA by default, is not something i'd argue for. In addition to being, as @loewenstein said, 3+ replicas you'd also expect anti-affinity in there and that's rather hard to have good configs for given all possible cluster topologies.

linsun:
if our default profile is for production usage, it does make sense to have replicas 2 for each control plane components.   who would run with 1 replica for each of istio control plane component in prod?  (I'm not sure if citadel can handle 2 replicas in 1.2, it can't in 1.1).

loewenstein:
@linsun as I have pointed out earlier, for production use we probably want a minimum of 3 replica and a PDB.

Otherwise the Istio control plane wouldn't be HA during node maintenance.

linsun:
Istio control plane isn't in the user's request flow (except mixer policy) so it should be okay if it is down during maintenance so I think down to 1 replica is a reasonable starting point during upgrade.

Lysholm:
I think this needs some more thought. These components shouldn't all be treated as equal as @linsun said.

The best example is certmanager, it takes certmanager between 10-15 seconds to start up and 1-5 minutes to process cert requests using letsencrypt and up to days with other issuers. Is the, worst case, 15% extra latency for cert issuance something you'd want to guard against? Maybe, but we're talking edge case scenario here. Probably not a candidate for disruption budget or HA by default.

The mixer and anything connected to the Adapter API is on the other end of the spectrum. Would i accept logging falling out during maintenance? Probably not.

Is it a goal that the default install should be ready for production out of box? In that case there's more work to be done. There's nothing stopping k8s from scheduling all mixer pods on the same node and in the same zone with the current config. So a single node failure can still take everything down and with the popularity of spot/preemtible nodes this is actually not that unlikeley.

loewenstein:
@linsun well, while Pilot is down any freshly scheduled pod would be blocked on Envoy waiting for config, wouldn't it?

Maybe that's an edge case, as multiple things have to go wrong to trigger failure, but node maintenance is exactly one of those scenarios where a couple of Pod reschedules are pretty likely...

howardjohn:
> @linsun well, while Pilot is down any freshly scheduled pod would be blocked on Envoy waiting for config, wouldn't it?
> 
> Maybe that's an edge case, as multiple things have to go wrong to trigger failure, but node maintenance is exactly one of those scenarios where a couple of Pod reschedules are pretty likely...

You also need the sidecar injector (and maybe citadel?) for new pods, as an extra data point

loewenstein:
@howardjohn @linsun Pilot was an example. Sure we need injector and probably Gateways. While I do think it is unfortunate that Citadel is currently a singleton, it is probably the least problematic with regard to HA - not sure if SDS changes that...

Lysholm:
What about adding new global options like this:
```
global.HA = {
  controlplane: {
    enabled: false,
    podAntiAffinity: {
       preferredTopologyKeys: [],
       requiredTopologyKeys: [],
    }
  },
  mesh: {
    enabled: true, 
    podAntiAffinity: {
      preferredTopologyKeys: ["failure-domain.beta.kubernetes.io/zone"]
      requiredTopologyKeys: ["kubernetes.io/hostname"]
    }
  }
```

enabling HA for mesh, enables PDBs and replicas on components that affect traffic in the mesh, while controlplane enables it on components required to make changes to the mesh.
The podAntiAffinity could be naively configured to do the best it could without knowing the cluster topology:

```
podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - weight: 100
            podAffinityTerm:
              labelSelector:
                matchExpressions:
                  - key: app
                    operator: In
                    values:
                    - istio-ingress
              topologyKey: "failure-domain.beta.kubernetes.io/zone"
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
                matchExpressions:
                  - key: app
                    operator: In
                    values:
                    - istio-ingress
            topologyKey: "kubernetes.io/hostname"
```
It still doesn't give regional HA. In the case zone-1 goes down while maintenance on zone-2 has all nodes down there then all istio-ingress pods will be rescheduled in zone-3. It's the best i can come up with without requiring user config. This would be a large change, since antiAffinity like this increases scheduling load.

rvandegrift:
The underlying issue can prevent cluster autoscaler from removing unutilized nodes.  I'm running istio 1.0 from the helm chart.  The singleton istio control plane pods happen to be on under utilized nodes, but CA cannot evict them due to the PDB.  So the node pool will never scale down.

This is an unfortunate default, and has been very difficult to track down.  I'd be very happy to have the option to require at least 2 replicas when a PDB could prevent autoscaling.  Alternatively, a flag to disable the PDBs could workt too.

IMO, this is a separate issue from enabling solid HA deployments from the helm charts.

Lysholm:
@rvandegrift: Such a flag exists already: `global.defaultPodDisruptionBudget.enabled = false`. The only other change that could be done for 1.1 or 1.0 is to flag this as a known issue in the documentation. Any change of default behavior here would be a breaking change.

rvandegrift:
> @rvandegrift: Such a flag exists already: `global.defaultPodDisruptionBudget.enabled = false`.

Oh thanks - this does exist in 1.1, but not 1.0 so I hadn't noticed.  That's a pretty good reason for us to upgrade.

> The only other change that could be done for 1.1 or 1.0 is to flag this as a known issue in the documentation. Any change of default behavior here would be a breaking change.

Fair enough, though I'd think there's still value in fixing a future release.  I haven't gone through the history to understand the current defaults, but I wonder how many folks would choose the existing situation over defaults with a working cluster autoscaler.

gateways/istio-egress/templates/autoscale.yaml
```diff
@@ -10,7 +10,15 @@ metadata:
     release: {{ .Release.Name }}
 spec:
   maxReplicas: {{ $gateway.autoscaleMax }}
+{{- if $gateway.autoscaleMin }}
   minReplicas: {{ $gateway.autoscaleMin }}
```
morvencao: Seem still an issue here, what if `autoscaleMin` is 1 and the actual replica is 1, in that case, the default PDB(with minAvailable=1 ) still prevents node drain.

gateways/istio-egress/templates/autoscale.yaml
```diff
@@ -10,7 +10,15 @@ metadata:
     release: {{ .Release.Name }}
 spec:
   maxReplicas: {{ $gateway.autoscaleMax }}
+{{- if $gateway.autoscaleMin }}
```
morvencao: Seem still an issue here, what if autoscaleMin is 1 and the actual replica is 1, in that case, the default PDB(with minAvailable=1 ) still prevents node drain.

gateways/istio-ingress/values.yaml
morvencao: Yes, the default PDB will prevent k8s node drain if the replica if selected pod is 1, so we need to be careful of HPA and PDB combination. There was an issue for cert-manager, the default replica is 1 while minAvailable of PDB is also 1, that would prevent node upgrade/maintain... 

Maybe we need clear doc(at least inline comments) to end users about this.

morvencao:
@rvandegrift PDB is introduced Istio from v1.1, there is flag to disable the PDBs, but the spec(minAvailable: 1) can't be customized: https://github.com/istio/istio/blob/release-1.1/install/kubernetes/helm/istio/values.yaml#L370-L375

istio-control/istio-config/values.yaml
ayj: cc @ozevren

ozevren: Have you tested this?

Lysholm: @ozevren: I'm running 3 replicas in production and nothing has critically failed atleast. Degredation or edge cases below my "detection threshold" might still be in there. Anything you want me to look for specifically?

costinm:
Sorry, post vacation and long thread, what is the plan for this PR ?

I believe we MUST have both a 'demo' profile with low resource as well as 2-3 easy 'prod' installs, with
reasonable defaults.  And at least one of the profiles needs to have a proper 2-3 replicas for each component, and be tested in the stability cluster.

gateways/istio-egress/templates/autoscale.yaml
costinm: Can we just document that if pod disruption budget is set, min must be set to at least 2 ?
It seems like a lot of complexity - normal k8s configs just work.

Keep in mind we want to start using kustomize as well - where we don't have all the template magic.

istio-control/istio-config/values.yaml
costinm: I did test galley with multiple replicas.

costinm: But we need to confirm the stability cluster runs multiple replicas of everything - I tested manually long ago.

Lysholm:
@costinm: That is fair. Closing this, this is not how i would do it or want istio setup. My current istio setup scripts set `global.defaultPodDisruptionBudget.enabled = false`. And apply the following PDBs afterwards: 

```
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  labels:
    app: istio-egressgateway
    istio: egressgateway
    release: istio
  name: istio-egressgateway
  namespace: istio-system
spec:
  maxUnavailable: 35%
  selector:
    matchLabels:
      app: istio-egressgateway
      istio: egressgateway
      release: istio
---
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  labels:
    app: istio-ingressgateway
    istio: ingressgateway
    release: istio
  name: istio-ingressgateway
  namespace: istio-system
spec:
  maxUnavailable: 35%
  selector:
    matchLabels:
      app: istio-ingressgateway
      istio: ingressgateway
      release: istio
---
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  labels:
    app: policy
    istio: mixer
    istio-mixer-type: policy
    release: istio
  name: istio-policy
  namespace: istio-system
spec:
  maxUnavailable: 35%
  selector:
    matchLabels:
      app: policy
      istio: mixer
      istio-mixer-type: policy
      release: istio
---
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  labels:
    app: telemetry
    istio: mixer
    istio-mixer-type: telemetry
    release: istio
  name: istio-telemetry
  namespace: istio-system
spec:
  maxUnavailable: 35%
  selector:
    matchLabels:
      app: telemetry
      istio: mixer
      istio-mixer-type: telemetry
      release: istio
```

Those four components are installed with the following helm values:

```
    autoscaleMin: 3
    podAntiAffinityLabelSelector:
    - key: app
      operator: In
      values: $ISTIO-COMPONENT
      topologyKey: "kubernetes.io/hostname"
    podAntiAffinityTermLabelSelector:
    - key: app
      operator: In
      values: $ISTIO-COMPONENT
      topologyKey: "failure-domain.beta.kubernetes.io/zone"
```

That gets me HA on all components in the request path while conserving resources on the control plane components and most importantly no interference with node maintenance.

PR #225, opened 2019-06-03
Update the gateway configs for telemetry add-ons

richardwxn:
Update the add-on gateway template and value.yaml
Remove ingress for grafana


gateways/istio-ingress/values.yaml
```diff
@@ -32,6 +32,18 @@ gateways:
       name: http2
     - port: 443
       name: https
+    - port: 15029
```
howardjohn: I think we had issues/complaints about too many gateway ports. Do these need to be here?

sdake: @howardjohn to maintain backward compatibility, we need to keep these ports intact, until such time as we provide a migration path.  These gateway ports expose the various dashboard addons.  To get to backward compatibility, this doc would need to be revised:

https://istio.io/docs/tasks/telemetry/gateways/

We should wait to merge said document until after 1.2 ships.  @howardjohn can you file an issue in istio.io on this topic and assign to @douglas-reid?j

For now, we should merge this section as is.

howardjohn:
@sdake opened at https://github.com/istio/istio.io/issues/4275

PR #226, opened 2019-06-04
added batch reporting parameters as helm variables

huang195:
Synchronize https://github.com/istio/istio/pull/14301 to this repo

### DIFF
--- a/istio-control/istio-discovery/values.yaml
+++ b/istio-control/istio-discovery/values.yaml
@@ -98,7 +98,7 @@ pilot:
 
 ## Mixer settings
 mixer:
-  telmetry:
+  telemetry:
     # Set reportBatchMaxEntries to 0 to use the default batching behavior (i.e., every 100 requests).
     # A positive value indicates the number of requests that are batched before telemetry data
     # is sent to the mixer server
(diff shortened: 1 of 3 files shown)


huang195:
@linsun can you please take a look at this PR since you reviewed the other PR that went into `istio/istio`?

istio-control/istio-discovery/values.yaml
```diff
@@ -94,4 +94,17 @@ pilot:
     # Will not define mixerCheckServer and mixerReportServer
     enabled: false
 
-  useMCP: true
\ No newline at end of file
+  useMCP: true
+
+## Mixer settings
+mixer:
```
linsun: seems having the top level mixer config in here is odd... mixer has its own config in https://github.com/huang195/installer/tree/tunnable_batch_report/istio-telemetry

huang195: right, if it's possible to move these to `istio-telemetry/mixer-telemetry`, that would be ideal. However, the configmap where these variables are needed is located in `istio-control/istio-discovery`. These are 2 different subcharts. Afaik, subcharts in helm don't share variable scope even with their parent chart.

huang195: Looking at it more closely, it appears they are not even subcharts, but independent charts. Can variables be shared across charts?

howardjohn: I believe this is the only way to have this work. when a user is configuring things, they will probably have one custom values.yaml that they share for all the charts so it won't be weird from a UX perspective

PR #233, opened 2019-06-07
Pull request validation pipeline

kramerul:
We created a pull request validation pipeline for concourse.


.concourse/set-pipeline.sh
```diff
@@ -12,4 +12,5 @@ if [ -n "$1" ] ; then
     fly --target $TARGET login --concourse-url $1
 fi
 
-fly -t $TARGET set-pipeline -c concourse.yaml -p $PIPELINE
\ No newline at end of file
+fly -t $TARGET set-pipeline -c concourse.yaml -p $PIPELINE
+fly -t $TARGET set-pipeline -c validate-pr.yaml -p validate-pr
```
c0d1ngm0nk3y: nit: we should rename the script to `set-pipelines.sh`

.concourse/validate-pr.yaml
```diff
@@ -0,0 +1,77 @@
...
+# git - can also push
+
+resource_types:
+  - name: pull-request
+    type: docker-image
+    source:
+      repository: teliaoss/github-pr-resource
+
+resources:
+#  - name: istio-installer #DO NOT SET TO FORK - for testing use another pipeline and pause this one
```
c0d1ngm0nk3y: This is probably just an artifact from copying, right?

.concourse/validate-pr.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+  - name: istio-installer-pr
+    type: pull-request
+    source:
+        repository: "istio/istio-installer"
+        access_token: ((GITHUB_ACCESS_TOKEN))
+
+jobs:
+  - name: validate-istio-installer-pr
+    serial_groups: [ cluster ]
```
c0d1ngm0nk3y: I think `serial_groups` only work within a pipeline [see here](https://github.com/concourse/concourse/issues/2691). So we should go with one pipeline or use the `pool` resource.

.concourse/validate-pr.yaml
```diff
@@ -0,0 +1,77 @@
...
+        status: pending
+    - task: validate
+      params:
+        KUBECONFIG: ((KUBECONFIG))
+      config:
+        platform: linux
+        image_resource:
+          type: docker-image
+          source:
+            repository: gcr.io/peripli/istio-base
```
c0d1ngm0nk3y: Should we really use an image from Peripli within istio? Maybe we rather use an istio image or push a new one if there is no suitable one.

.concourse/validate-pr.yaml
```diff
@@ -0,0 +1,77 @@
...
+        inputs:
+          - name: istio-installer-pr
+          - name: istio-master
+        run:
+          path: /bin/sh
+          args:
+          - -ec
+          - |
+            cd istio-installer-pr
+            # HUB=istionightly TAG=nightly-master make run-build
```
c0d1ngm0nk3y: still needed?

PR #235, opened 2019-06-12
Update kustomize templates

ConcourseWdf:
This PR is created automatically!


jkbschmid:
@geeknoid @howardjohn 
Is there a way to give this bot clearance for the CLA-agreement?

jkbschmid:
Side note: Install-demo fails because the make-target creates an invalid test/demo/k8s.yaml, so this failure is kind of '_intended_' for this PR.

geeknoid:
No sorry, that's against corporate policy because it could be used to subvert the CLA process and sneak in some non-CLAed changes. So PRs generated by this bot will need manual review.

Could this be run as part of the build process instead?

howardjohn:
@geeknoid the reason we have the generated files in the repo is so users can do things like `kubectl -k github.com/istio/install/...` and install Istio in one command

geeknoid:
It's fine to have the generated files checked in. But can these just be
checked in as part of a human doing a local build and commit, like we do in
the API repo?

c0d1ngm0nk3y:
How to continue with this PR? I see 4 options

1) To keep it a manual process to sync those files (not really an option IMHO)
2) Use a CI to update those files automatically (like proposed in this PR), but opens some questions with the CLA bot.
3) To integrate it in the build process (make), so that it is at least hard to get those files out-of-sync.
4) Do not checkin generated files in the first place. @howardjohn I see your use case with the `one command install`, but would't it be an option to generate and add those files in the release process? That sounds more natural to me.

I would prefer option 4. @howardjohn Do you think that could be an option?

geeknoid:
Assuming generating those files isn't a terrible burden, not checking in
generated files is definitely preferable.

holgero:
@c0d1ngm0nk3y If we go for option 4 (generate the files in a release build), it would be no longer a one step installation from master. OTOH, I am not sure if we really need that option (of a one-step-installation from current master). Maybe it is even better this way, so that customers are directed to the released versions. And for developers it should be a reasonable request, that they have to generate the files by themselves first.

PR #236, opened 2019-06-12
Automatically create PRs to update kustomize templates

jkbschmid:
In order to keep `kubectl apply -k` in sync with installing istio via iop/helm, we have created a pipeline job that automatically builds the installer and creates a PR to istio/installer with the updated kustomize templates.

An exemplary PR is https://github.com/istio/installer/pull/235.

### DIFF
--- a/.concourse/concourse.yaml
+++ b/.concourse/concourse.yaml
@@ -123,9 +123,9 @@ jobs:
         run:
           path: /bin/sh
           args:
-          - -ec
+          - -ecu
           - |
-            git config --global credential.helper 'store
+            git config --global credential.helper store
             echo "https://ConcourseWdf:${CONCOURSE_WDF_PASSWORD}@github.com" > ~/.git-credentials
             set -x
             cd istio-installer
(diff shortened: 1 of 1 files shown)


sdake:
@jkbschmid i've been a bit out of the loop on the installer.  Hope to fix that soon.  I don't know much about concourse ci, other than its a ci system :).  Could you expand a bit on what @costinm agreed to related to this repo and concourse ci?

Cheers
-steve

c0d1ngm0nk3y:
> Could you expand a bit on what @costinm agreed to related to this repo and concourse ci?

As far as I know, the idea was to setup an initial update pipeline that can be used for CI. So that we can contribute to the installer without the fear of breaking it. But in the meantime, it came out that there is no real update process. So the pipeline is very short at the moment - install and test :)

Currently, the pipeline runs on `postsubmit`, but it is already in progress to use it for PR validation as well. The concourse is currently maintained by a SAP team, the pipeline is of course checked in.

We thought it might be good idea to generate those kustomize files. Because it is always tricky to have generated files within the VCS. So we used the same CI server for this task.

@holgero Right?

@sdake What would be an alternative for such jobs?

holgero:
@c0d1ngm0nk3y Exactly.

holgero:
As stated in https://github.com/istio/installer/pull/235#issuecomment-502701011 and in https://github.com/istio/installer/pull/235#issuecomment-505485984 it seems to make more sense to generate the kustomize templates in a release build only. Therefore I will close this PR, too.

costinm:
One alternative: we can add a test, and require that any developer runs "make test" before sending a PR. 
( and make sure "make test" is updating the templates ). Benefit is that we'll verify make test was run as well..

holgero:
@costinm 
> One alternative: we can add a test, and require that any developer runs "make test" before sending a PR.
> ( and make sure "make test" is updating the templates ). Benefit is that we'll verify make test was run as well..

Sounds good. I think we can do that. So we would create a validation job that verifies that the templates in the PR are correct.

PR #237, opened 2019-06-13
Remove presubmit label

clyang82:
The installer code changes cannot pass the ci due to https://github.com/istio/istio/pull/14680. remove the `presubmit` in this repo accordingly.

### DIFF
--- a/test/tests.mk
+++ b/test/tests.mk
@@ -118,7 +118,7 @@ run-test.integration.kube.presubmit:
 
 	set -o pipefail; \
 	cd ${GOPATH}/src/istio.io/istio; \
-	${GO} test -v ${INT_TARGETS} --istio.test.select +customsetup ${INT_FLAGS} 2>&1 | tee ${GOPATH}/out/logs/$@.log
+	${GO} test -v ${INT_TARGETS} --istio.test.select -customsetup,-flaky ${INT_FLAGS} 2>&1 | tee ${GOPATH}/out/logs/$@.log
 
 run-stability:
 	 ISTIO_ENV=${ISTIO_CONTROL_NS} bin/iop test stability ${GOPATH}/src/istio.io/tools/perf/stability/allconfig ${IOP_OPTS}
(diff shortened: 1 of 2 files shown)


test/tests.mk
```diff
@@ -118,7 +118,7 @@ run-test.integration.kube.presubmit:
 
 	set -o pipefail; \
 	cd ${GOPATH}/src/istio.io/istio; \
-	${GO} test -v ${INT_TARGETS} --istio.test.select -customsetup,+presubmit ${INT_FLAGS} 2>&1 | tee ${GOPATH}/out/logs/$@.log
+	${GO} test -v ${INT_TARGETS} --istio.test.select +customsetup ${INT_FLAGS} 2>&1 | tee ${GOPATH}/out/logs/$@.log
```
howardjohn: This should be `-customsetup,-flaky`. These are labels, and we are saying we want to skip the ones that need custom setup and that are flaky. Before we said skip customsetup ones and only do ones with "presubmit" label

clyang82: Thanks for your comment. I am a little confusion about that. before the change, this will select tests that have ```label.Presubmit``` only. It will **not** select tests that have both ```label.CustomSetup```
and ```label.Presubmit```.

the oz's change is 
> Replacing the use of presubmit with customsetup in tests and examples

so in my understand, we should select test that have ```label.Customsetup``` only.

howardjohn: Before: `!customSetup && presubmit`
Now: `!customSetup && !flaky`

The presubmit label has essentially been replace with the flaky label (which is the inverse of the presubmit label)

Oz's comment was misleading here. What he meant was that before they had some tests to test the label mechanism. For this test, they used Presubmit label, but it could have been any arbitrary label. So since presubmit was removed, he chose an arbitrary one, customsetup

Hope that makes sense

clyang82: @howardjohn Thanks for your explanation. It is clear now for me.

PR #241, opened 2019-06-18
Use strict version for zipkin

clyang82:
There has some problems I think
1. The image is from docker hub.
image
if we use the version `2`,  it may lead our tested version is not same as released version. for example: when we did test for 2.14.1 and client may use 2.14.2 for their environment.

2. we have another PR to ensure the `imagePullPolicy` is `IfNotPresent`. after that, the latest or our desired image cannot be applied to client's environment if there is an old version `2` exists.

FYI @morvencao 

### DIFF
--- a/istio-telemetry/tracing/values.yaml
+++ b/istio-telemetry/tracing/values.yaml
@@ -36,7 +36,7 @@ tracing:
 
   zipkin:
     hub: docker.io/openzipkin
-    tag: 2
+    tag: 2.14.2
     probeStartupDelay: 200
     queryPort: 9411
     resources:


morvencao:
Should we also get this to istio main repo? I'm thinking that istio 1.1.x and 1.2.x will also need this.

clyang82:
> Should we also get this to istio main repo? I'm thinking that istio 1.1.x and 1.2.x will also need this.

Yes. I think we should fix this in 1.1.x and 1.2.x repo

linsun:
@clyang82 have  you tested this version 2.14.2?  If you can validate that, I will merge it and please propose a fix for master and 1.2.x also.  If you need it in 1.1.x ping @duderino

clyang82:
> @clyang82 have you tested this version 2.14.2? If you can validate that, I will merge it and please propose PRs for master and 1.2.x also. If you need it in 1.1.x ping @duderino

@linsun Yes. I have verified the version 2.14.2, it works normally. Thanks.

PR #246, opened 2019-06-20
update image version to latest stable one

morvencao:


istio-control/istio-autoinject/templates/mutatingwebhook.yaml
```diff
@@ -35,8 +35,6 @@ webhooks:
         operator: NotIn
         values:
         - disabled
-      - key: istio-env
```
howardjohn: If you remove this I'll merge. But I don't think we want this, see discussion on https://github.com/istio/installer/pull/247

howardjohn:
@morvencao I think a lot of these changes already landed in master, can you rebase so we can see whats changed + the tests pass?

morvencao:
@howardjohn Looks like all [changes](https://github.com/istio/installer/blob/master/global.yaml#L38-L41) have already in master. I'm going to close this.

PR #247, opened 2019-06-20
Sidecar auto-inject with istio-env label

clyang82:
Here is the description in readme
> Only one auto-injector environment should have enableNamespacesByDefault=true, which will apply that environment to any namespace without an explicit istio-env label.

I think we should use the same label for injecting all namespaces or selected namespaces. before my change,
```
istio-bookinfo-1-32720    Active   116m   istio-env=istio-control,istio-injection=enabled,istio-testing=istio-test
```
istio-bookinfo-1-32720 cannot have sidecar injected if `enableNamespacesByDefault` is true.

### DIFF
--- a/istio-control/istio-autoinject/templates/mutatingwebhook.yaml
+++ b/istio-control/istio-autoinject/templates/mutatingwebhook.yaml
@@ -36,7 +36,7 @@ webhooks:
         values:
         - disabled
       - key: istio-env
-        operator: DoesNotExist
+        operator: Exists
 {{- else }}
       matchLabels:
         istio-env: {{ .Release.Namespace }}


clyang82:
@costinm in this PR https://github.com/istio/istio/pull/13706 add `istio-env ` label. the current behaviour in circleci is that cannot inject sidecar for 
```
istio-bookinfo-1-40066    Active   79m   istio-env=istio-control,istio-injection=enabled,istio-testing=istio-test
istio-bookinfo-1-48569    Active   82m   istio-env=istio-control,istio-injection=enabled,istio-testing=istio-test
reachability-1-93264      Active   77m   istio-env=istio-control,istio-injection=enabled,istio-testing=istio-test
```

istio-control/istio-autoinject/templates/mutatingwebhook.yaml
```diff
@@ -36,7 +36,7 @@ webhooks:
         values:
         - disabled
       - key: istio-env
-        operator: DoesNotExist
+        operator: Exists
```
jkbschmid: Is there a reason why this entry is not removed? If injection is enabled by default, istio-env should not matter IMO.

linsun: I think it is for multi istio control plane scenarios which is supported in the new installer.

The change makes sense to me, @costinm @howardjohn WDYT?

howardjohn: Im not sure why we have the matchLabels immediately below this and this part as well? will need to read the docs and see if this makes sense I guess

howardjohn: oh its an `if else` nevermind...

I think we either need this change or need to remove the block -- not sure which is better. @costinm ?

howardjohn: I think the problem is when you have two injectors. If you don't have this, then `enableNamespacesByDefault` means you can only have one injector as it will match on everything.

Not sure what the best solution is, maybe we should just not do the labeling in the test or not have this by default

costinm: If a user explicitly set an istio-env label - it means they want to select a specific version, maybe a canary or test version.

This is pretty critical - otherwise there is no way to test a new version of injector when 'enableByDefault' is set.

costinm: Also I think the change would pretty much break the feature - it's an 'and' between match expressions, so namespaces without any label will be ignored. 

Was there any testing done on this PR ? We absolutely need to add a test with 'enableNamespacesByDefault' ( or drop the feature if we are not testing it )...

clyang82: from the client perspective, I think if istio installed with `enableNamespacesByDefault=true`, then in the following scenario the sidecar should be injected automatically except for the namespace with `istio-injection=disabled`
```
default           Active   10m     <none>
istio-control     Active   8m51s   istio-injection=disabled
istio-ingress     Active   7m39s   istio-injection=disabled
istio-policy      Active   7m16s   istio-injection=disabled
istio-system      Active   8m56s   istio-injection=disabled
istio-telemetry   Active   7m33s   istio-injection=disabled
kube-node-lease   Active   10m     <none>
kube-public       Active   10m     <none>
kube-system       Active   10m     <none>
istio-test1       Active   10m     istio-env:istio-control,istio-injection=enabled
istio-test2       Active   10m     istio-injection=enabled
istio-test2       Active   10m     istio-env:istio-control
```
if the namespace with label `istio-env:istio-control,istio-injection=enabled` cannot have sidecar injected, that is a surprise.
so I think we should remove the block by default
```
      - key: istio-env
        operator: DoesNotExist
```

if the client want to have 2 injectors, they can manually modify the `matchExpressions`, maybe we can have a document to guide client how to do it. @costinm WDYT?

costinm: In your example - istio-test2 should be injected, using istio-control injector. 

Same for istio-test1 - why would it be a surprise to have it injected if istio-injection==enabled ? 

The whole point of 'istio-env' is to support multiple injectors and configs - so users can test their 
injector config changes without breaking everything. 

We clearly expect zero or one 'default injectors', but the default injectors should skip any namespace that 
has 'istio-injection=disabled' OR explicitly selects a specific injector ( canary, custom config, etc).

costinm: To be clear: I agree that in your example a sidecar should be injected in all cases except those 'disabled'.
But I disagree that a namespace where "istio-env:istio-control' is set should be injected with the default
injector - it explicitly specifies a config, it would be surprising to user to not have the specified profile configured.

clyang82: That makes sense for me.  Thanks for your explanation.  I think I need to fix in some test cases to have `enableNamespacesByDefault` passed to decide it should have `istio-env` labels defined or not.

costinm:
I mean: if there are still issues we can discuss in env WG and based on resolution close the PR or move ahead.

clyang82:
The current solution is OK for me. I have another PR in istio/istio to be merged to follow the current logic. Thanks.

PR #248, opened 2019-06-20
Update the clusterrole for citadel

clyang82:
cannot create configmap `istio-citadel` without proper permission.


howardjohn:
This is consistent with istio/istio, not adding new permissions.

PR #253, opened 2019-06-24
make all images configurable

anandkumarpatel:
Some image names have been hardcoded in helm templates.
This PR makes all the image names configurable via helm values.

I have defaulted them to original values in their respective values.yaml to ensure backward compatibility.

why?
In my use-case, all images have to come from an internal image repository and it has strict naming conventions. This change allows changing image names without having to maintain a fork and change image names in the source.

### DIFF
--- a/istio-cni/templates/daemonset.yaml
+++ b/istio-cni/templates/daemonset.yaml
@@ -48,7 +48,7 @@ spec:
         # This container installs the Istio CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: {{ .Values.cni.hub }}/install-cni:{{ .Values.cni.tag }}
+          image: {{ .Values.cni.hub }}/{{ .values.cni.image | default "install-cni" }}:{{ .Values.cni.tag }}
           imagePullPolicy: {{ .Values.cni.pullPolicy | default .Values.global.imagePullPolicy }}
           command: ["/install-cni.sh"]
           env:
--- a/istio-telemetry/prometheus-operator/templates/prometheus.yaml
+++ b/istio-telemetry/prometheus-operator/templates/prometheus.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     release: {{ .Release.Name }}
 spec:
-  image: "{{ .Values.prometheus.hub }}/prometheus:{{ .Values.prometheus.tag }}"
+  image: "{{ .Values.prometheus.hub }}/{{ .Values.prometheus.image | default "prometheus" }}:{{ .Values.prometheus.tag }}"
   version: {{ .Values.prometheus.tag }}
   retention: {{ .Values.prometheus.retention }}
   scrapeInterval: {{ .Values.prometheus.scrapeInterval }}


istio-cni/templates/daemonset.yaml
```diff
@@ -48,7 +48,7 @@ spec:
         # This container installs the Istio CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: {{ .Values.cni.hub }}/install-cni:{{ .Values.cni.tag }}
+          image: {{ .Values.cni.hub }}/{{ .values.cni.image }}:{{ .Values.cni.tag }}
```
costinm: Please add default values ( with the old values ).

istio-cni/values.yaml
```diff
@@ -1,5 +1,6 @@
 cni:
   hub: gcr.io/istio-release
+  image: install-cni
```
costinm: And maybe remove this - the default values ( | default ...)  will be used. 

I don't think this is a common use case - the official release uses the default names, and advanced users who build  their  own can still override the setting. 

IMO using Kustomize or something similar would be a cleaner approach.

costinm (commented):
I'm not very sure this is the right solution - operator and kustomize can patch the image to arbitrary value, 
I was hoping we can simplify this in the templates. But at least add the default value.

costinm:
I'm curious, what are the 'strict naming conventions' ? As you can see, our naming conventions are not very strict, would be good to improve them.

anandkumarpatel:
All of our assets (repo name, docker images, helm charts ...etc) have to be of the form <product>-<component>-<asset>. 

This helps our automation scrips and integrations.

istio-cni/values.yaml
anandkumarpatel: Yeah I would like to use kustomize as well but our system right now is strictly helm 😅.

Will update to move default into the chart though.

anandkumarpatel: done

istio-cni/templates/daemonset.yaml
anandkumarpatel: done

costinm:
https://github.com/istio/istio/issues/15264

costinm:
Sorry - can you give an example ? <product>-<component>-<asset> - is product=istio, component=pilot 
and asset= ???

I'm asking because our naming convention is all over the place - 'nodeagent' is my favorite example ( should be istio-sdsagent or similar ). I was going to propose we make image and container names more consistent, happy to see what people are commonly using.

Added a feature request for the larger problem of 'image configuration'

costinm (approved):
Approved this PR - it just makes those components a bit more consistent with the others.

anandkumarpatel:
Thanks!
Here are some examples
product=istio, component=pilot and asset= <this is optional. Say if `pilot` had a `ui` and `cli`>

costinm:
Please resolve conflicts and let's get it merged. 

I think I get the point - you have one hub and in it images from various vendors. 
It would make sense to have all docker images for istio start with istio- and have the component
consistently as second part. 

We generally didn't do this because the repo is 'istio' - so istio/istio-pilot would be redundant, but didn't consider the use case of 'vendoring' the images.

Probably the name should be istio-control-pilot, istio-security-citadel... If anyone wants to make naming consistent - I wouldn't say no...

anandkumarpatel:
@costinm sorry for the delay! rebased and it should be good to go now.

anandkumarpatel:
rebased @costinm looks like the other changes were integrated by someone else. These are the ones that are left.

howardjohn (approved):

PR #257, opened 2019-06-25
Update SDS in config map

JimmyCYJ:
Protect the SDS config fields by helm flag global.sds.enabled (https://github.com/istio/istio/pull/15109), and keep the SDS config consistent with istio/install/kubernetes/helm/istio/templates/configmap.yaml


JimmyCYJ:
Thank you @howardjohn

JimmyCYJ:
https://github.com/istio/istio/issues/13439

PR #258, opened 2019-06-26
Add helm linting to the repository

sdake:
This adds helm linting to the repository.  There are several lint failures
which will be fixed in a followup PR.  The main objective of this work is to
start transitioning from circleci to prow.  In order to do that, I need to
understand how prow is configured.


Makefile
```diff
@@ -159,6 +159,9 @@ kind-run:
 		${KIND_CLUSTER}-control-plane \
 		bash -c "cd ${GOPATH}/src/istio.io/installer; make ${TARGET}"
 
+# Runs helm lint on the various charts
+lint:
```
howardjohn: this target already exists, not sure how make will handle this

sdake: I rebased - must have missed htis.  Will address.

prow/lib.sh
```diff
@@ -0,0 +1,161 @@
...
+
+function setup_and_export_git_sha() {
+  if [[ -n "${CI:-}" ]]; then
+    if [[ "${CI:-}" == 'bootstrap' ]]; then
+      # TODO: Remove after update to pod-utils
+
+      # Make sure we are in the right directory
+      # Test harness will checkout code to directory $GOPATH/src/github.com/istio/istio
+      # but we depend on being at path $GOPATH/src/istio.io/istio for imports
+      if [[ ! $PWD = ${GOPATH}/src/istio.io/istio ]]; then
```
howardjohn: not sure if this will work as is or needs to be `/installer`. the test isn't running here so hard to tell

prow/istio-lint.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
costinm: Can you hold on this until we know what the new Prow setup and version will look like ? 
My understanding is that they may support a proper yaml configuration - tektoncd or similar. 

I am very uncomfortable with all those scripts - we have a clear goal to use Kind not Mason, 
and there all kind of downloads that don't make any sense - make test is using docker images and 
has download support for stuff that is needed. 

Ok to use Prow - but we should have a clean solution and put all that is common across CI/CDs in make.

prow/lib.sh
```diff
@@ -0,0 +1,161 @@
...
+    unsetup_clusters
+  fi
+  if [[ "${USE_MASON_RESOURCE}" == "True" ]]; then
+    mason_cleanup
+    cat "${FILE_LOG}"
+  fi
+}
+
+# Set up a GKE cluster for testing.
+function setup_e2e_cluster() {
```
costinm: Please use KIND only. AFAIK the agreement is that testing will be done in KIND.

prow/lib.sh
```diff
@@ -0,0 +1,161 @@
...
+
+function check_kind() {
+  echo "Checking KinD is installed..."
+  if ! kind --help > /dev/null; then
+    echo "Looks like KinD is not installed."
+    exit 1
+  fi
+}
+
+function setup_kind_cluster() {
```
costinm: There is a make target.

prow/lib.sh
```diff
@@ -0,0 +1,161 @@
...
+  # Create KinD cluster
+  if ! (kind create cluster --name=e2e-suite); then
+    echo "Could not setup KinD environment. Something wrong with KinD setup. Please check your setup and try again."
+    exit 1
+  fi
+  KUBECONFIG="$(kind get kubeconfig-path --name="e2e-suite")"
+  export KUBECONFIG
+}
+
+function cni_run_daemon() {
```
costinm: This should be in makefile and tests - no reason it can't be done using 'make test' and only be run in prow.

sdake: re `prow/lib.sh` - the thinking is this file prow/lib.sh will end up in the `pkg` repo (https://github.com/istio/pkg) as a common file for all to use.  I didn't alter this file - its simply a copy from istio/istio and I expect we will just sync up when the file hits the pkg repo.

prow/lib.sh
sdake: see comment later on prow/lib.sh.  We can optimize our version of lib.sh, however, I think it makes much more sense s a project to standardize the prow library code - even if it results in unused duplication.

prow/lib.sh
sdake: yes, the PR to add this to prow only uses the kind image.  We probably need to have a wider discussion with T&R or the TOC about how to standardize this prow/lib.sh script across the various repositories.

prow/istio-lint.sh
sdake: I can hold off - I just wanted to start learning about "old prow" prior to making the switch to "new prow" so the switch can move quickly.

As a project there is a strong TOC desire to eliminate the dependency on circleci.

prow/lib.sh
sdake: I am not using most of this prow/lib.sh code :)

sdake:
placing on hold for a week or two until new prow gets sorted out.  If new prow takes longer then a few weeks, we are going to transition the repo to prow so we have plenty of time to stabilize any flakey operation of the tests.

prow/lib.sh
sdake: Also, if my last comment wasn't clear, kind will be used for prow work.

howardjohn:
@sdake this is failing because we are using helm `&version.Version{SemVer:"v2.7.2", GitCommit:"8478fb4fc723885b155c924d1c8c410b7a9444e6", GitTreeState:"clean"}` on the builder image and it doesnt support `-f` in lint

sdake:
@howardjohn going to abandon this PR as after learning about prow, this implementation isn't consistent with our plans in T&R.

If you wish to take it over feel free to clone it.

PR #260, opened 2019-06-26
Fix noauth test flakes

howardjohn:


howardjohn:
@costinm one failure down, 3 to go...

PR #262, opened 2019-06-27
Turn on namespaces by default

howardjohn:
Currently the test framework can either always set the namespace label
or never set it. We need to make it smarter then we can turn this back
on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -299,7 +299,8 @@ jobs:
       - run:
           command: make sync
       - run:
-          command: make kind-run TARGET="install-base ENABLE_NAMESPACES_BY_DEFAULT=false"
+          # todo: set ENABLE_NAMESPACES_BY_DEFAULT=false once integration tests can support it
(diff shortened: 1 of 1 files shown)


howardjohn:
Only needed/depends on https://github.com/istio/istio/pull/15138

howardjohn:
@costinm got all tests passing except minikube. No clue why its failing.. it only fails on a pod calling itself with no sidecar. Shouldn't be istio related at all

howardjohn:
Looks like it is independent of the installer repo, so we are finally passing all tests the main repo passes. See https://github.com/istio/istio/issues/15222

costinm:
For multi-namespace install: we can add installer-specific tests, no need to run the full istio integration suite. 

Hopefully the 'conformance' test suite ( or a subset of istio integration tests ) will be usable in future, but as long as the stability/perf clusters run multiple control planes and we have basic pre-submit tests 
we should be good.

PR #264, opened 2019-07-01
cherrypick removing hardcoded istio-system

quanjielin:
cherrypick change in https://github.com/istio/istio/pull/15199 to installer

### DIFF
--- a/security/nodeagent/templates/daemonset.yaml
+++ b/security/nodeagent/templates/daemonset.yaml
@@ -44,6 +44,10 @@ spec:
           {{- end }}
             - name: "Trust_Domain"
               value: "{{ .Values.global.trustDomain }}"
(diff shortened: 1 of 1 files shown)


costinm:
Can you update the change description to include what it does and why ? I thought namespace 
is already available as an environment variable defined by k8s.

quanjielin:
updated description to put Istio PR link, the change on yaml is borrowed from pilot - https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/charts/pilot/templates/deployment.yaml#L98

costinm:
Never mind, found the info, updated the title. 
/lgtm  - not very happy with using NAMESPACE when injector uses POD_NAMESPACE, but maybe we'll change the injector env instead for consistency...

PR #146, opened 2019-05-09
Update inject-template.yaml

crazyxy:
Backport changes from https://github.com/istio/istio/pull/13801

PR #265, opened 2019-07-01
Update builder to kind v0.4.0

howardjohn:
I will move to a real istio repo if this passes tests ok, adding hold for now

### DIFF
--- a/test/docker/Dockerfile
+++ b/test/docker/Dockerfile
@@ -13,16 +13,16 @@ FROM docker:latest as docker
 ############################################
 # Image to create go binaries
 FROM golang:1.12.5 as golang
-ENV     GO111MODULE=on
-ENV     GOPROXY=https://proxy.golang.org
+ENV GO111MODULE=on
 
-# IMPORTANT: kind releases may or may not work with the 1.14.1 kindest/node tag.
-# We want to create images with multiple k8s versions.
-# Use latest since there is not a specific release to match the current kindest/node:v1.14.1 image
-RUN GO111MODULE="on" go get -u sigs.k8s.io/kind@master
+# IMPORTANT: kind releases may or may not work with other kindest/node images.
+# When updating, make sure you update the kindest/node version as well
+# The KinD release notes will include the safe images to use
+RUN go get -u sigs.k8s.io/kind@v0.4.0
 
 # Used to upload test results to test grid
-RUN go get -u istio.io/test-infra/toolbox/ci2gubernator
+# TODO add -u flag back to ci2gubernator once it works correctly. For now it is failing
+RUN go get istio.io/test-infra/toolbox/ci2gubernator
 RUN go get -u github.com/jstemmer/go-junit-report
 
 # get helm binary - do here to limit to only binary in final image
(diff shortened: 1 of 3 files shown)


test/docker/Dockerfile
```diff
@@ -13,16 +13,16 @@ FROM docker:latest as docker
 ############################################
 # Image to create go binaries
 FROM golang:1.12.5 as golang
-ENV     GO111MODULE=on
-ENV     GOPROXY=https://proxy.golang.org
```
costinm: Why did you remove the proxy ? I think it was supposed to speed things up and make sure deleted repos don't affect us.

test/docker/Dockerfile
```diff
@@ -35,12 +35,12 @@ RUN chmod +x /usr/local/bin/repo
...
 # create istioctl from `master`, assuming that's what the builder had locally
 # ignores `The command '/bin/sh -c go get -d istio.io/istio' returned a non-zero code: 1`
-RUN go get -d istio.io/istio || true   
+RUN git clone https://github.com/istio/istio /go/src/istio.io/istio
 RUN cd /go/src/istio.io/istio && make istioctl
 
 ############################################
 # Main image
-FROM  kindest/node:v1.14.1
+FROM kindest/node:v1.15.0@sha256:b4d092fd2b507843dd096fe6c85d06a27a0cbd740a0b32a880fe61aba24bb478
```
costinm: They don't have a tag ?

howardjohn: `v1.15.0` is the tag. In the release notes the recommend adding the sha I guess its not needed though

test/docker/Dockerfile
howardjohn: `build istio.io/test-infra/toolbox/ci2gubernator: cannot load github.com/google/go-github/github: unexpected status (https://proxy.golang.org/github.com/google/go-github/@v/v0.0.0-20171108000855-8c08f4fba5e0.zip): 400 Bad Request` if it is included. Not sure why

howardjohn:
This will fix https://github.com/istio/installer/pull/146/files because the docker image has istioctl built into it so it was old. @qfel

test/docker/Dockerfile
costinm: Did it fail once or consistently ? We should check what is the speed impact, and maybe look for a different proxy.

test/docker/Dockerfile
costinm: I really don't like SHAs copied by hand. It's the job of tools.

test/docker/Dockerfile
howardjohn: It is failing every time. I tried 8 days ago, 4 days ago and today. Not sure why though.. either way this should only effect the docker image being built, so speed is not a major concern right?

test/docker/Dockerfile
howardjohn: I just removed it

howardjohn:
blocked by https://github.com/istio/istio/issues/15432

howardjohn:
Ah missed one.. Also blocked by https://github.com/istio/istio/pull/15463

test/docker/Dockerfile
ericvn: I just tried pulling it locally this morning using the `GOPROXY=https://goproxy.io` and that worked fine. I have had issues even getting the _istio/istio_ modules to download with _https://proxy.golang.org_ (the later is still a Beta I believe).

PR #267, opened 2019-07-02
Pass enableNamespacesByDefault to istio config

clyang82:
in this PR https://github.com/istio/istio/pull/15284, we enable get `enableNamespacesByDefault` from command line. pass this value in the new installer to enable test with `enableNamespacesByDefault` is false

Signed-off-by: clyang82 <clyang@cn.ibm.com>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -299,8 +299,7 @@ jobs:
       - run:
           command: make sync
       - run:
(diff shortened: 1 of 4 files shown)


howardjohn:
I turned of the test that sets this, can you turn it back on ?Line 302 of .circleci/config.yaml

clyang82:
@howardjohn done. I need to have the istio/istio#15284 merged firstly, then monitor if circleci can be passed or not.

PR #268, opened 2019-07-02
Add PR validation to check if generated files are still in sync

c0d1ngm0nk3y:
Additionally update the kustomize templates via `make run-build` to be up-to-date again.

We noticed to make this work, the technical github user @ConcourseWdf would need some authorizations, probably commenting. But definitely *NOT* commiting.


phil9909:
I'm ok with my commits being contributed to this project.

c0d1ngm0nk3y:
@linsun I have brought up this question already here (https://github.com/istio/istio/issues/14687#issuecomment-505903378). Is it possible to get such authorizations for the technicl user to be able to update the status on the PR. Alternatively we would need an ACCESS_TOKEN of another technical user with such authorizations.

costinm:
IMHO: a bot that generates periodically would be a better long term solution.

We might keep the generated files in a different branch - or even a different repo.

We generate the artifacts to make it super easy to install with "apply -k" which currently requires git.

PR #269, opened 2019-07-02
specify enable namespaces by default in demo

howardjohn:
test/demo/k8s.yaml is autogenerated

This will fix the installer tests on istio/istio


howardjohn:
Blocked by https://github.com/istio/istio/issues/15309. We probably need to rethink our injection selection logic

howardjohn:
@costinm can you take a look?

kustomize/citadel/namespace.yaml
```diff
@@ -2,3 +2,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: istio-system
+  labels:
+    name: istio-system
```
costinm: Why ?

howardjohn: Otherwise our check for 
```
      - key: name
        operator: NotIn
        values:
        - {{ .Release.Namespace }}
```

Does not work

howardjohn: `key:` only looks at labels

costinm: I mean - why not adding injection=disabled.

howardjohn: sgtm, should we just remove the name check then since it isn't documented to add `name` and we don't add name

costinm: probably - but not a priority to remove it. 
I've seen other apps ( certmanager ) use the same pattern with foo-injection=disabled.

PR #271, opened 2019-07-03
Make minikube not required for merging

howardjohn:
The test will still run, just not block merges. The same tests are run in KinD so I am not too concerned about losing coverage


clyang82:
the minikube issue was fixed in https://github.com/istio/installer/pull/272. I think we can leave as it is.

PR #275, opened 2019-07-10
fix missing crds for certmanager.

morvencao:
Sync the cert-manager crds changes from istio-operator https://github.com/istio/operator/pull/69

Although the CI for new installer apply all crds with `kubectl`, it's necessary add crds for cert-manager in chart template, because the istio-operator renders all the CRDs with crds chart.

### DIFF
--- a/crds/templates/crds.yaml
+++ b/crds/templates/crds.yaml
@@ -3,5 +3,5 @@
 {{ .Files.Get "files/crd-12.yaml" }}
 {{- if .Values.certmanager.enabled }}
 {{ .Files.Get "files/crd-certmanager-10.yaml" }}
-{{ .Files.Get "files/crd-certmanager-10.yaml" }}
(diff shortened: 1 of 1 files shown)


crds/templates/crds.yaml
```diff
@@ -1,3 +1,7 @@
 {{ .Files.Get "files/crd-10.yaml" }}
 {{ .Files.Get "files/crd-11.yaml" }}
 {{ .Files.Get "files/crd-12.yaml" }}
+{{- if .Values.certmanager.enabled }}
+{{ .Files.Get "files/crd-certmanager-10.yaml" }}
+{{ .Files.Get "files/crd-certmanager-10.yaml" }}
```
howardjohn: same file twice?

morvencao: oh...  have updated the second to 11.

PR #277, opened 2019-07-12
add dns certs for pilot in dual control plane.

morvencao:


costinm:
That doesn't mean I'm holding this PR - just starting a discussion... If it helps having this merged to unblock some use case - I don't mind, but I would rather use 'proper cert' or SPIFFEE and not continue to use the citadel-generated DNS certs which have little benefit but are hard to configure ( restart citadel since it's a CLI flag, etc)

morvencao:
Is the migration from DNS cert to SPIFEE SAN for pilot finished?
 Because I found that the `append-dns-names` is still `true`, see: https://github.com/istio/installer/blob/master/security/citadel/templates/deployment.yaml#L38

costinm:
No harm in merging it, just to be safe.

The bootstrap-generated config checks for:
"spiffe://cluster.local/ns/ISTIO_NAMESPACE/sa/istio-pilot-service-account"

PR #279, opened 2019-07-12
Make demo directly install citadel

howardjohn:
Currently the demo will generate config for all components but then
reference this pregenerated config for citadel. This changes it to align
with other components and be self contained.

Ran make run-build-demo which is where most changes here come from

### DIFF
--- a/test/demo/k8s.yaml
+++ b/test/demo/k8s.yaml
@@ -12,7 +12,7 @@ metadata:
     istio: sidecar-injector
 data:
   values: |-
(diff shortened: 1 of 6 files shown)


howardjohn:
The hold stuff was just testing github - if someone actually approves this feel free to remove it

test/demo/k8s.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
     istio: sidecar-injector
```
costinm: This should probably move under 'kustomize/' directory, for consistency ( it was created before we started to pre-generate profiles ). Can be separate PR.

test/demo/values.yaml
```diff
@@ -3,6 +3,16 @@
 # Includes components used in the demo, defaults to alpha3 rules.
 # Note: please only put common configuration for the demo profiles here.
 global:
+  istioNamespace: istio-system
```
costinm: It would be really great if the demo used the split-namespace model. 
It's a demo, intended for fresh install/remove in non-prod clusters (since it takes over istio-system ) - no upgrade concerns.

test/demo/k8s.yaml
howardjohn: We could do this, but it would require to redefine the demo profile from being a set of values.yaml to kustomizations, right?

test/demo/values.yaml
howardjohn: Is this still the plan with the new same-namespace upgrade plan you made? Or would that be each component upgrades in the same namespace, but we still have split namespaces?

test/demo/k8s.yaml
costinm: Not really. We can provide one 'kustomization' for each profile - or just support a single profile, do we really need all the demo variants ?

test/demo/values.yaml
costinm: Let's go with the simplest - all in istio-system.

PR #281, opened 2019-07-15
Beginning the work to improve SDS setup 

costinm:
I'm breaking this into multiple PRs, for easier review.

This is renaming the deployment/etc - nodeagent is super confusing, and adds proper defaults 
and kustomize options so it can be installed a-la-carte as an optional standalone component. 

Tested a-la-carte manually using: 
``bash
kubectl apply -k crds
kubectl apply -k kustomize/citadel
kubectl apply -k kustomize/sds-agent
```

The automated test depends on several other components - will be run after all pieces are in ( some are in progress in istio/istio )

This PR also fixes some breakages on 'local' mode.

### CODE REVIEW

File: security/citadel/templates/deployment.yaml [hunk 1]
```diff
@@ -37,12 +37,6 @@ spec:
           args:
             - --append-dns-names=true
             - --grpc-port=8060
-          # global.tag may contain something like "release-1.1-latest-daily". Assume >1.2 if we cannot extract semver version.
-          {{- if regexFind "[0-9]+\\.[0-9]+\\.[0-9]+" .Values.global.tag | default "1.3.0" | semverCompare "<1.2" }}
-            - --grpc-hostname=citadel
-          {{- else}}
-            - --grpc-host-identities=citadel
```
howardjohn: :+1:, I had to do this for https://github.com/istio/istio/pull/15440 as well

security/nodeagent/templates/daemonset.yaml
```diff
@@ -1,47 +1,52 @@
...
-        app: istio-nodeagent
-        istio: nodeagent
+        app: istio-sdsagent
+        istio: sdsagent
         release: {{ .Release.Name }}
       annotations:
         sidecar.istio.io/inject: "false"
     spec:
-      serviceAccountName: istio-nodeagent-service-account
+      serviceAccountName: istio-sdsagent
```
howardjohn: whats the reason for this change? I think all other components have `-service-account`

security/nodeagent/templates/daemonset.yaml
```diff
@@ -1,47 +1,52 @@
...
 {{- else }}
-          image: "{{ .Values.global.hub }}/{{ .Values.nodeagent.image }}:{{ .Values.global.tag }}"
+          image: "{{ .Values.global.hub }}/{{ .Values.sdsagent.image }}:{{ .Values.global.tag }}"
 {{- end }}
           imagePullPolicy: {{ .Values.global.imagePullPolicy | default "Always" }}
           args:
           {{- if .Values.global.logAsJson }}
             - --log_as_json
           {{- end }}
+            - --log_output_level=all:debug
```
howardjohn: probably don't want this on by default?

security/nodeagent/values.yaml
```diff
@@ -1,16 +1,22 @@
...
+  # This can be used to specify a different certificate provider
+  #
+  # This replaces the env variable in 1.2 alpha - env variables are not meant to be exposed to users for supported
+  # configs.
+  caProvider: "Citadel"
+
+  # Address of the certificate provider server.
+  # By default the node agent is installed in same namesapce with citadel, so namespace not needed.
+  #
+  # Important: the citadel cert has hardcoded "citadel" and "citadel.NS.svc.cluster.local" - any provider will need
```
howardjohn: isn't it `istio-citadel`? and `istio-ca` will work too

test/sds/sds.mk
```diff
@@ -0,0 +1,102 @@
+
+
+# Target to run SDS graceful upgrade tests
+test-sds:
```
howardjohn: are you planning to add to circleci?

security/nodeagent/templates/daemonset.yaml
costinm: For no reason. It's a bit tricky to change for old components ( pilot in particular ), but at least we can stop doing it.

security/nodeagent/templates/daemonset.yaml
costinm: Right - I'm still testing. Will remove soon - but need to add some setting.

security/nodeagent/values.yaml
costinm: Right, forgot to update the comment.

test/sds/sds.mk
costinm: yes, but it's a long way to get there ( probably a week or so ) - some things are broken in agent.

costinm:
PTAL

test/sds/envoy-sds-template.yaml
```diff
@@ -0,0 +1,484 @@
...
+            "connect_timeout": "{{ .connect_timeout }}",
+            "lb_policy": "ROUND_ROBIN",
+            {{ if eq .config.ControlPlaneAuthPolicy 1 }}
+            "tls_context": {
+              "common_tls_context": {
+                "alpn_protocols": [
+                  "h2"
+                ],
+                {{ if 0 }}
+                {{ if 0 }}
```
howardjohn: i assume this is WIP?

costinm: Partially WIP - it's intended to test file-based SDS cert rotation, but that's manual.
I can remove it from the PR if you want.

istio-control/istio-discovery/templates/configmap-envoy.yaml
```diff
@@ -113,19 +109,46 @@ data:
           tls_context:
             require_client_certificate: true
             common_tls_context:
+              alpn_protocols:
+              - h2
+
               validation_context:
```
quanjielin: if sds enabled, the root cert is also fetched through sds (so here in validation_context needs to be sdsconfig), could be updated in later PR.

istio-control/istio-discovery/templates/configmap-envoy.yaml
```diff
@@ -113,19 +109,46 @@ data:
...
                 trusted_ca:
                   filename: /etc/certs/root-cert.pem
 
-              alpn_protocols:
-              - h2
-
+  {{- if .Values.global.sds.enabled }}
```
quanjielin: so later here and below ```target_uri```/ token ```filename``` will be read from opts["**"] after pilot agent change get in.

kustomize/sds-agent/sds-agent.yaml
```diff
@@ -0,0 +1,128 @@
...
+        - name: sdsagent
+          image: "gcr.io/istio-release/node-agent-k8s:master-latest-daily"
+          imagePullPolicy: Always
+          args:
+            - --log_output_level=all:debug
+          volumeMounts:
+            - mountPath: /var/run/sds
+              name: sdsudspath
+          env:
+            - name: CA_PROVIDER
```
quanjielin: could remove ```CA_PROVIDER``` and ```CA_ADDR``` from DaemonSet deployment (and leave ```values.yaml``` to set those values, so that nodeagent could be configured to connect to multiple CAs.

kustomize/sds-agent/sds-agent.yaml
```diff
@@ -0,0 +1,128 @@
...
+              value: Citadel
+            - name: CA_ADDR
+              value: istio-citadel:8060
+            - name: "Trust_Domain"
```
quanjielin: in istio/istio, trustdomain is global configmap/meshconfig and passed to by pilot/nodagent/citadel/etc, how new installer plan to handle trust domain ?

security/nodeagent/values.yaml
```diff
@@ -1,16 +1,22 @@
...
-    CA_ADDR: ""  
-    # names of authentication provider's plugins.
-    Plugins: ""
+sdsagent:
+
+  # This can be used to specify a different certificate provider
+  #
+  # This replaces the env variable in 1.2 alpha - env variables are not meant to be exposed to users for supported
+  # configs.
+  caProvider: "Citadel"
```
quanjielin: for citadel today, we need to setup ```VALID_TOKEN: true``` in env (since usenormaljwt==true in this PR, citadel haven't use trusworthy jwt), this will nodeagent keep envoy sds connection (without disconnect and expected an updated token in cert rotation)

security/nodeagent/templates/daemonset.yaml
```diff
@@ -1,47 +1,54 @@
...
             - mountPath: /var/run/sds
               name: sdsudspath
           env:
-          {{- if .Values.nodeagent.env }}
-          {{- range $key, $val := .Values.nodeagent.env }}
+            - name: CA_PROVIDER
+              value: {{ .Values.sdsagent.caProvider }}
+            - name: CA_ADDR
+              value: {{ .Values.sdsagent.caAddr }}
+          {{- if .Values.sdsagent.env }}
```
quanjielin: env "Trust_Domain" is needed (as kustomize folder does)

istio-control/istio-discovery/templates/configmap-envoy.yaml
costinm: It could be - but to minimize risks ( and take advantage of cert rotation and multi-cluster) we can also load it from the mounted file. 

This reduces the load ( less generated config ) and work ( since SDS doesn't currently support all the merging and custom roots )

istio-control/istio-discovery/templates/configmap-envoy.yaml
costinm: yes, waiting for that to get in.

kustomize/sds-agent/sds-agent.yaml
costinm: All other configs we ship work out-of-box and can be configured for users if the defaults are not what they want.

We also need this for 'kustomize' style.

kustomize/sds-agent/sds-agent.yaml
costinm: New installer doesn't have any 'globals' ( to reduce risks ), by design.
And the meshconfig is never exposed to user - all is from either values.yaml or kustomizations.

security/nodeagent/values.yaml
costinm: Wait, citadel doesn't support 'trustworthy JWT' ? I didn't test it because k8s didn't seem to support it.

security/nodeagent/templates/daemonset.yaml
costinm: Seems to be working.. I'll check

security/nodeagent/values.yaml
myidpt: Citadel supports trustworthy JWT.

myidpt:
Costin, do you have time to work on this? If not, we will take it over.

costinm:
Obsolete, istiod replaces it.

PR #283, opened 2019-07-18
remove namespace for meshpolicy.

morvencao:
remove namespace for cluster scoped resource: `MeshPolicy`.

### DIFF
--- a/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
+++ b/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
@@ -17,21 +17,6 @@ spec:
 
 {{- if .Values.global.mtls.enabled }}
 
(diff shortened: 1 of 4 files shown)


costinm:
Wait a minute, when did we end up with a cluster wide resource ? Is it alpha or beta ? Approved ? 

This has major implications on new installer - the main point is to be able to make safe upgrades, without a global change that can bring down the entire cluster.

costinm:
I would rather remove the 2 MeshPolicies from installer - and make it clear users should probably not use them, and if they do it's not installer's fault. 

/hold

morvencao:
@costinm The `meshPolicy` api is still in alpha, not sure if these changes will block upgrade, I can give it a try.

costinm:
I think the best option is to try to get rid of the MeshPolicy from new installer. If user wants to use them - they can, but the goal is to avoid 'cluster wide' / 'global' changes.

PR #284, opened 2019-07-18
Sync charts with istio/operator

ostromart:
Mostly adding missing separators. Only non-trivial change is the creation of the namespace.


crds/templates/crds.yaml
```diff
@@ -1,3 +1,12 @@
+# To prevent accidental injection into istio-system.
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Release.Namespace }}
```
howardjohn: we may want to add all the namespaces here not just .Release.Namespace, can be in the future though

ostromart: That's a good idea - added.

costinm: Please don't add namespaces to a file names 'crds...'

ostromart: Done.

crds/templates/namespaces.yaml
```diff
@@ -0,0 +1,64 @@
+# To prevent accidental injection into istio control plane namespaces.
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Release.Namespace }}
+  labels:
+    istio-operator-managed: Reconcile
```
costinm: Unrelated  to this PR: but the pattern should be to specify which operator is managing this namespace ( if that's what we want ). 

"Reconcile" is quite confusing.

ostromart: I've created https://github.com/istio/istio/issues/15825 to track this.

PR #287, opened 2019-07-25
Add sidecars for kiali

clyang82:
Add the missed sidecars in kiali clusterrole.yaml file.


linsun:
This LG to me.  I recall you had another PR that has this similar change but in upgrading kiali to v1.1.

PR #290, opened 2019-07-26
Starting to add version and isolate deployments in same namespace

costinm:
This is the start of supporting multiple isolated versions of Istio in same namespace, and removing the need to create a new namespace to isolate.

For 1.3 we are reducing the scope to pilot + injector, eventually other components will be added.

The mechanism is similar with knative - the config is bundled with the deployment.

The main trick is separating cluster-wide resources from the deployment+config and adding version.


istio-control/istio-autoinject/templates/mutatingwebhook.yaml
```diff
@@ -1,16 +1,17 @@
+{{ if not .Values.skipWebhook }}
```
howardjohn: can you add this to values.yaml if it is intended to be used? Not sure why someone would want the injector without a webhook, unless there is an existing one pointing to it already?

istio-control/istio-autoinject/templates/serviceaccount.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.clusterResources }}
```
howardjohn: Is there any use case to enable this rather than installing with the CRDs?

istio-control/istio-config/templates/poddisruptionbudget.yaml
```diff
@@ -15,6 +15,6 @@ spec:
       app: galley
       release: {{ .Release.Name }}
       istio: galley
+      version: {{ .Values.version }}
```
howardjohn: `version` has no default, will this break? Because I think this will template to `version: ` and `name: istio-galley-`.

test/inplace/inplace.mk
```diff
@@ -0,0 +1,7 @@
+
+# Installs Istio 1.2 using default install yaml, install canary, verify canary starts.
+test-inplace:
+	kubectl apply -k kustomize/istio-1.2/default --prune -l release=istio
+	kubectl apply -k kustomize/istio-canary --prune -l release=istio-canary
```
howardjohn: won't the prune here delete the existing install?

istio-control/istio-autoinject/templates/mutatingwebhook.yaml
costinm: This is used internally by the installer - I don't think it's something that we want users to actually do. 

Installer and operator may install the injector without a webhook, and use some tool to create the secret and inject the webhook manually later.

Or even create a webhook only ( without the autoinjector ) and point it to a deployment on a different cluster. 

Also webhook manipulation requires very high cluster-wide permissions, while upgrading the injector and deployments does not. 

In general I am trying to decouple the (deployment+config+service) from additional cluster wide resources ( like the webhook ).

istio-control/istio-autoinject/templates/serviceaccount.yaml
costinm: Old installer and most of our tests are generating those cluster resources - not ready to disable it yet. 

However in future I think it is clearly better to separate the cluster resources ( and the high-priv credentials needed to deploy them ) from the actual deployment+config. A CI/CD may push upgrades to various components - but it would only need istio-system ( or istio-telemetry ) permissions, not cluster-wide rights.

istio-control/istio-config/templates/poddisruptionbudget.yaml
costinm: I added version ('default') to global.yaml.

I'm not entire sure what's the best option - I can add 'if' and some extra to add the '-' where needed, to preserve the old behavior ( of not having a version ).

test/inplace/inplace.mk
costinm: No - prune is looking for any resource with 'release=istio-canary' ( and there will be none in this case ) and if it doesn't exist in the new template it'll remove it.

If later we deploy again istio-canary, with an option that removes one of the deployments - when prune is called it will delete the missing deployment.

PR #291, opened 2019-07-28
add configurable rolling update strategy.

morvencao:
Context: https://github.com/istio/istio/issues/13552 & https://github.com/istio/istio/issues/15878

### DIFF
--- a/gateways/istio-egress/templates/deployment.yaml
+++ b/gateways/istio-egress/templates/deployment.yaml
@@ -20,6 +20,10 @@ spec:
     matchLabels:
       app: istio-egressgateway
       istio: egressgateway
+  strategy:
+    rollingUpdate:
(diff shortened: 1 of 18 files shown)


gateways/istio-egress/templates/deployment.yaml
```diff
@@ -20,6 +20,10 @@ spec:
     matchLabels:
       app: istio-egressgateway
       istio: egressgateway
+  strategy:
```
costinm: Should this be protected by an if as well ? Doesn't seem to be something we want in 100% of cases.

linsun: why wouldn't?  For demo cases we don't care about upgrade?

costinm: Good point. We actually do care probably, but no reason it wouldn't work.

PR #299, opened 2019-07-31
Mixer should use local useMCP var and Mixer's sidecar should be configurable

huang195:
This PR addresses 2 issues:
- Mixer is currently hard coded to run with a sidecar. This should be controllable similarly to Pilot.
- Mixer is current using `global.useMCP`, but Pilot is using `pilot.useMCP`. We should be consistent, so I introduced a `mixer.telemetry.useMCP`.

### DIFF
--- a/istio-telemetry/mixer-telemetry/templates/deployment.yaml
+++ b/istio-telemetry/mixer-telemetry/templates/deployment.yaml
@@ -87,7 +87,7 @@ spec:
 {{- if .Values.global.logAsJson }}
           - --log_as_json
 {{- end }}
-{{- if .Values.mixer.telemetry.useMCP }}
+{{- if .Values.global.useMCP }}
     {{- if .Values.global.controlPlaneSecurityEnabled}}
           - --configStoreURL=mcp://localhost:15019
     {{- else }}
@@ -122,7 +122,7 @@ spec:
 {{ toYaml .Values.global.defaultResources | indent 10 }}
 {{- end }}
         volumeMounts:
-{{- if .Values.mixer.telemetry.useMCP }}
+{{- if .Values.global.useMCP }}
         - name: istio-certs
           mountPath: /etc/certs
           readOnly: true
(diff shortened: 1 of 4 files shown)


istio-telemetry/mixer-telemetry/templates/deployment.yaml
```diff
@@ -87,7 +87,7 @@ spec:
 {{- if .Values.global.logAsJson }}
           - --log_as_json
 {{- end }}
-{{- if .Values.global.useMCP }}
+{{- if .Values.mixer.telemetry.useMCP }}
```
costinm: At least respect global - if global is false it typically means MCP is not installed. 

Any reason to have a telemetry-specific setting ?

istio-telemetry/mixer-telemetry/values.yaml
```diff
@@ -27,6 +27,8 @@ mixer:
         cpu: 4800m
         memory: 4G
 
+    useMCP: true
```
costinm: And we must stop adding uncommented settings ( and in general add settings only if there are strong use cases for them, it is already far too much )

istio-telemetry/mixer-telemetry/templates/deployment.yaml
huang195: Pilot uses a local `useMCP` here: https://github.com/istio/installer/blob/master/istio-control/istio-discovery/templates/configmap.yaml#L119. I was trying to make the variable being used in Pilot and Mixer more consistent, but if the recommended way is to remove local `useMCP` and stick with the global one, I'm fine with that as well.

huang195:
OK, let me know if you'd rather have me remove the use of Mixer's local `useMCP` and change Pilot's `useMCP` to a global var

huang195:
I went ahead and made the change described above

huang195:
@costinm friendly ping to see what else is needed here?

istio-telemetry/mixer-telemetry/templates/deployment.yaml
costinm: I think it's better as you have it, globals are bad.
( and besides they are no longer really global - each component is now a separate module )

huang195: ok, let me revert the changes so that we avoid using globals.

huang195:
@costinm can you take a quick look as we had a few back and forth on using global vs local here. Just want to make sure this is what you had in mind as well.

PR #301, opened 2019-08-01
add security namespace.

morvencao:
Merge https://github.com/istio/operator/pull/139 back to new installer.


costinm:
I'm going to close this, please reopen if there is some use cases or reason to not keep security in istio-system. 

Feedback on too many namespaces has been quite negative.

PR #302, opened 2019-08-01
disable scrap endpoints that require certificate when prometheus.security=false

huang195:
When prometheus.secure is set to false, prometheus deployment will not mount from the istio.default secret. However, it will still try to scrap from the `kubernetes-pods-istio-secure` endpoint that requires the certificate mounted from istio.default secret, resulting in unnecessary error messages in prometheus pods. This disables the scrap endpoint when prometheus.security=false

### DIFF
--- a/istio-telemetry/prometheus-operator/templates/servicemonitors.yaml
+++ b/istio-telemetry/prometheus-operator/templates/servicemonitors.yaml
@@ -114,6 +114,7 @@ spec:
       action: replace
       targetLabel: pod_name
 ---
+{{- if .Values.prometheus-operator.security.enabled }}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
@@ -169,6 +170,7 @@ spec:
     - sourceLabels: [__meta_kubernetes_pod_name]
       action: replace
       targetLabel: pod_name
+{{- end }}
 ---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
@@ -215,6 +217,7 @@ spec:
       action: replace
       targetLabel: pod_name
 ---
+{{- if .Values.prometheus-operator.security.enabled }}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
@@ -270,6 +273,7 @@ spec:
     - sourceLabels: [__meta_kubernetes_pod_name]
       action: replace
       targetLabel: pod_name
+{{- end }}
 ---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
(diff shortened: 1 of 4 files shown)


huang195:
I'm looking at `prometheus-operator`, and its behavior is a bit different from `prometheus`'s helm chart. In `prometheus`, it uses a flag `security.enabled` to indicate if `Citadel` is enabled. When it is enabled, it creates a secret named `istio.default`, from which various certificate files can be generated. These are then used to scrap `https` endpoints. However, in `prometheus-operator`, there's no equivalent flag (I can certainly add one), but it uses a different secret called `istio.prometheus`. I'm checking with @douglas-reid to see what is the intention of a different name.

huang195:
@costinm the changes that were made to `prometheus`'s Helm chart to better handle the case where Citadel is disabled have been synchronized to `prometheus-operator`. AFAIK, no functionality is broken without this PR. Prometheus still does scraping as expected. It's just kind of annoying to see prometheus's log file getting filled with error messages about certificate files not being found every 15 seconds (default scrape interval)

istio-telemetry/prometheus-operator/templates/servicemonitors.yaml
```diff
@@ -114,6 +114,7 @@ spec:
       action: replace
       targetLabel: pod_name
 ---
+{{- if .Values.prometheus-operator.security.enabled }}
```
clyang82: `.Values.prometheus-operator.security.enabled` should be `.Values.prometheus.security.enabled`. it breaks the `make docker-run-test TEST_TARGET="run-prometheus-operator-config-test"`

huang195: I actually added a `.Values.prometheus-operator.security.enabled` parameter as part of this PR, so it should be self-contained.

huang195: @clyang82 so I'm looking at test scripts, and it looks like we're not installing `prometheus-operator` like the other components, i.e., using `bin/iop` from our own helm charts. Instead we're installing it by `curl` from a `coreos`'s repo, see here: https://github.com/istio/installer/blob/master/test/install.mk#L185. This PR unfortunately exposed this problem. Let me know what you think.

clyang82: we assume the prometheus-operation should be pre-installed in client's real environment. so here https://github.com/istio/installer/blob/master/test/install.mk#L185 is just install prometheus-operation crd.
for the cr installation, refer to https://github.com/istio/installer/blob/master/test/install.mk#L201. I think the problem is https://github.com/istio/installer/blob/master/istio-telemetry/prometheus-operator/values.yaml#L1 so in the whole file, we use `.Values.prometheus.xxx`.

huang195: yeah, you're right. Let me fix that

PR #303, opened 2019-08-02
Change to fit for container run as nonrun

clyang82:
Resolve the leftover issue for nonroot user to run container. refer to https://github.com/istio/istio/issues/15984

### DIFF
--- a/istio-control/istio-config/templates/deployment.yaml
+++ b/istio-control/istio-config/templates/deployment.yaml
@@ -43,7 +43,7 @@ spec:
 {{- end }}
           imagePullPolicy: {{ .Values.global.imagePullPolicy | default "Always" }}
           ports:
-          - containerPort: 443
+          - containerPort: 9443
           - containerPort: 15014
           - containerPort: 15019
           - containerPort: 9901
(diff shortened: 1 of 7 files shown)


clyang82:
The three failed tests were fixed in https://github.com/istio/istio/pull/16014

istio-control/istio-config/templates/deployment.yaml
```diff
@@ -66,6 +66,7 @@ spec:
           - --validation-webhook-config-file
           - /etc/config/validatingwebhookconfiguration.yaml
           - --monitoringPort=15014
+          - --validation-port=9443
```
costinm: Does the flag exists in 1.2 ? Upgrade.

costinm:
Force merge since everything is badly broken without this

istio-control/istio-config/templates/deployment.yaml
clyang82: This flag is just temporarily fix in installer. it should be not required when this PR is merged - https://github.com/istio/istio/pull/16015. You can see that there does not get the default value wile just hardcoded.
https://github.com/istio/istio/blob/8757a56331a384bae999dbdbb165eea6175ffd60/galley/cmd/galley/cmd/server.go#L160

PR #305, opened 2019-08-06
Cleanup unused resources, regenerate

costinm:
As discussed, in 1.3 new installer will use istio-system and in-place upgrade, with the exception 
of providing a canary pilot in istio-system and support knative 'side-by-side' installation.

This removes the unsupported kustomizations - the addition of 'istio-cluster' provides more 
flexibility on deployment without using multiple namespaces, and in 1.4 we expect some 
components will be replaced ( mixer -> wasm, prometheus use its own operator, etc).


kustomize/default/gen-istio-ingress.yaml
```diff
@@ -102,6 +102,9 @@ spec:
       labels:
         app: istio-ingressgateway
         istio: ingressgateway
+        heritage: Tiller
```
howardjohn: is this expected?

istio-control/istio-autoinject/values.yaml
```diff
@@ -24,6 +24,10 @@ sidecarInjectorWebhook:
   nodeSelector: {}
   tolerations: []
 
+  # If set, will use the value as injection label. The value must match the 'release' label of the injector,
+  # except when 1.2 istio-injection label is used, which must be set to "enabled".
+  injectLabel: istio-injection
```
howardjohn: If we do this don;t we need to replace `  istio-injection: enabled` with `{{injectLabel}}: enabled` on L64 of webhook?

PR #306, opened 2019-08-06
Update common files.

geeknoid:


Makefile
```diff
@@ -1,321 +1,52 @@
...
+# figure out all the tools you need in your environment to make that work.
+export BUILD_WITH_CONTAINER ?= 0
+ifeq ($(BUILD_WITH_CONTAINER),1)
+IMG = gcr.io/istio-testing/build-tools:2019-08-05
+UID = $(shell id -u)
+PWD = $(shell pwd)
+GOBIN_SOURCE ?= $(GOPATH)/bin
+GOBIN ?= /work/out/bin
+
+RUN = docker run -t --sig-proxy=true -u $(UID) --rm \
```
costinm: I don't think it'll work for kind - it creates files under /home/USER/.kube (the kubeconfigs to be used later steps )

Makefile
```diff
@@ -1,321 +1,52 @@
...
+	-v /etc/passwd:/etc/passwd:ro \
+	-v $(readlink /etc/localtime):/etc/localtime:ro \
+	--mount type=bind,source="$(PWD)",destination="/work" \
```
costinm: Please bind destination to PWD as well. Otherwise symlinks and other things are odd

Makefile
```diff
@@ -1,321 +1,52 @@
...
+	--mount type=volume,source=istio-go-mod,destination="/go/pkg/mod" \
+	--mount type=volume,source=istio-go-cache,destination="/gocache" \
+	--mount type=bind,source="$(GOBIN_SOURCE)",destination="/go/out/bin" \
+	-w /work $(IMG)
```
costinm: Did you test this ? How can it work without priviledged ? The tests are creating a kind cluster, AFAIK it still needs priv.

And since the base image is different and creates other containers - not sure this is the most efficient solution

Makefile
sdake: I think we will need more time to integrate the work in the installer with common.  Kind makes this work especially complex.

Makefile
sdake: this has been tested only for non-kind operational modes.  For this particular repository the common implementation will need work.

sdake: > Did you test this ? How can it work without priviledged ? The tests are creating a kind cluster, AFAIK it still needs priv.
> 
> And since the base image is different and creates other containers - not sure this is the most efficient solution

Costin,
For more context we are changing out the base image currently in use in CI/CD with the image at https://gcr.io/istio-testing/build-tools which is built from this dockerfile: https://github.com/istio/tools/blob/master/docker/build-tools/Dockerfile.  We have generated a common build system that requires only docker and make in order to operate and uses pinned versions of all tools for repeatable build and testing.  In the case of the above linked dockerfile, there are some deficiencies in the dockrefile.  Kind is missing.  bazel is missing.

The install repo has gone to great lengths to optimize around kind and we don't wish to disturb that.  This functionality defaults to off, although personally, I wouldn't merge the work until this common work is functional with this repo (proxy is in a similar situation).

Cheers
-steve

Makefile
costinm: Well - I think common must first be able to test and work well to kind, since this has implications on how the base image looks like, etc.

We already run everything in installer using a docker image, including testing - and it's clearly 
possible and a good idea, but if 'common' wants to reinvent/duplicate this it should do it properly :-)

Makefile
costinm: Well - the install repo IS using a base image that can not only build everything go-related in a docker container, but it can also efficiently run Kind tests ( and soon multi-cluster kind tests). 

So either istio-testing is using the same image as install, or they create a better one - but I don't think it should regress, regardless if it is on by default or not.

costinm:
How about: start with the image and makefile from istio/installer, and improve it if something is missing  instead of starting from scratch and not supporting Kind.

geeknoid:
@sdake @costinm I believe you're missing the point of this global update to the common files, and are over-estimating the durabiity of it all.

All I'm really doing is rolling out the new structure for the files (Makefile vs Makefile.core.mk). The fact is that unless you set the BUILD_WITH_CONTAINER environment variable, the build proceeds exactly  as it did before. We have  plenty of opportunity to improve the common files further over time as we enable the new container-based builds across all repos.

costinm:
While I like env variables for CLI ( and kustomize ) - for the installer
makefile things are more complicated and will be even more so as we add
multicluster tests.

I'm not sure even the structure is right... Not happy with what we use now
- but maybe a more incremental approach would work better. Clearly good to
have common targets and helpers - maybe we can have a target that clones
the common repo an include helper .mk files ? And leave the main Makefile
for real targets ( that's a pain point with current makefile in installer)

Makefile
sdake: thanks @costinm - will take a look at how installer does the work so we have a common framework across repos.

Makefile
sdake: the "base image" is vague :).  We have a few kicking around, and we are working on making just one to deal with that is smaller but has the necessary tools.  If I had to do over again I would have studied your work in more detail.  There is still time to take this path :)

The build dockerfile however, IMO is a writeoff.  That is the part we are reinventing from first principles.  I hear you around not breaking your workflows 👍

howardjohn:
One thing I do like about the common-files approach:

In the installer today we have targets like `run-foo` and `foo`. Some run in the container and some don't, it isn't super consistent. With the common files approach we have the same targets for container or not, and just call them with `make foo` or `make -f Makefile.core.mk foo` to decide to use container or not. Its a bit more wordy to use ` -f Makefile.core.mk` I guess, bu its more consistent

geeknoid:
Or you can set the BUILD_WITH_CONTAINER variable to 0 or 1 to control
whether to use the container or not, and then just always use the top-level
Makefile and ignore Makefile.core.mk.

PR #309, opened 2019-08-08
Expose the option --enable-reconcileWebhookConfiguration

lei-tang:
Expose the option --enable-reconcileWebhookConfiguration, which allows a user to turn on/off Galley managing its own webhookconfiguration.
Related PR: https://github.com/istio/istio/pull/16118

[ ] Docs
[X ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure


rlenglet:

PR #310, opened 2019-08-08
Add the option --enableManageWebhookConfig

lei-tang:
Add the option --enableManageWebhookConfig to allow turning on/off Sidecar-Injector managing its own webhookconfiguration.

Related PR: istio/istio#16121

[ ] Docs
[X ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/istio-control/istio-autoinject/templates/deployment.yaml
+++ b/istio-control/istio-autoinject/templates/deployment.yaml
@@ -49,7 +49,7 @@ spec:
             - --injectConfig=/etc/istio/inject/config
             - --meshConfig=/etc/istio/config/mesh
             - --port=9443
-            - --enableSidecarInjectorToManageWebhookConfig={{ .Values.sidecarInjectorWebhook.enableSidecarInjectorToManageWebhookConfig }}
+            - --enableReconcileWebhookConfiguration={{ .Values.sidecarInjectorWebhook.enableReconcileWebhookConfiguration }}
             - --healthCheckInterval=2s
             - --healthCheckFile=/health
                   {{- if eq .Release.Namespace "istio-system"}}
--- a/istio-control/istio-autoinject/templates/deployment.yaml
+++ b/istio-control/istio-autoinject/templates/deployment.yaml
@@ -49,7 +49,7 @@ spec:
             - --injectConfig=/etc/istio/inject/config
             - --meshConfig=/etc/istio/config/mesh
             - --port=9443
-            - --enableManageWebhookConfig={{ .Values.sidecarInjectorWebhook.enableManageWebhookConfig }}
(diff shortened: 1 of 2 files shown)


costinm (commented):
This doesn't make sense: there is a webhook that is added anyways. 

You want to add a flag around the webhook itself (mutatingwebhook.yaml). 
And the flag should be simpler, like

webhokEnable: true

The sidecar injector is only patching the cert - unlike galley.

lei-tang:
@costinm This flag is not about enabling or disabling webhook. Instead, it is about enabling/diabling that a webhook configures its own webhookconfiguration. A webhook having the privilege to configure its webhookconfiguraiton is a security risk.

costinm:
Right now we have a lot of cluster-wide configs that are applied. 

We do need a way to separate cluster-wide settings and permissions - but not by adding fine-grained configs, will confuse the user. 

The plan is to have a single option ( clusterResources) that will prevent the generation of any 
cluster-wide resource or config. You can check if clusterResources==false, and use that to configure components to not generate any cluster-wide config change.

costinm:
BTW - separating the cluster-wide configs is a proposal that still needs to be reviewed. 

As it is, the PR doesn't work: if you set this flag istio won't work. I would love to see cluster-wide separated, but we can't add options that don't work. Try to install with the option enabled - or add 
a test for the option. 

There are quite a few other options that are likely broken and not tested - but this doesn't mean
we can keep adding, in particular if it's clear they don't work.

lei-tang:
A higher level config option does not necessarily conflict with fine-grained config options: they each have their own pros and cons. An option of finer granularity offers more flexibility: a user may want to control a specific component instead of multiple components. Both higher level config options and finer granular options are useful and offer more choices. 

The default value of the config option in this PR is true, which is the same as the behavioir in the current Sidecar Injector implementation, i.e., Sidecar Injector configures its webhookconfiguration.
That is to say, this PR will not affect the working of Sidecar Injector. When Istio Webhook Controller is deployed, a user can set this config option as false to stop Sidecar Injector configuring its webhookconfiguration.

costinm:
I think we need to consider usability - values.yaml is already considered too complex and unfriendly.

We don't have use cases - or test capacity - for all permutations of options.
As I mentioned, the final solution is not ready yet - missing component will also add several other setttings. 

We also have the choice of  having a single option to indicate that DNS certs are managed by 
another component ( to control citadel generation of DNS certs ), and using the 'cluster-globals' to
control components installation/management of cluster-wide resources.

rlenglet:

PR #312, opened 2019-08-08
Expose the option --append-dns-names

lei-tang:
Expose the option --append-dns-names, which sets the CLI parameter append-dns-names of Citadel: 
- If true, Citadel will generate webhook certificates.
- If false, Citadel will not generate webhook certificates.
If there is another component generate webhook certificates, this parameter should be set as false.

The matching PR on istio/istio: https://github.com/istio/istio/pull/16142

[ ] Docs
[X ] Installation
[ ] Networking
[ ] Policies and Telemetry
[X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/security/citadel/templates/deployment.yaml
+++ b/security/citadel/templates/deployment.yaml
@@ -38,7 +38,7 @@ spec:
             {{- if .Values.global.sds.enabled }}
             - --sds-enabled=true
             {{- end }}
-            - --append-dns-names={{ .Values.security.appendDnsNames }}
+            - --append-dns-names={{ .Values.security.enableGenerateWebhookCerts }}
             - --grpc-port=8060
             - --citadel-storage-namespace={{ .Release.Namespace }}
           {{- if .Values.kustomize }}
--- a/security/citadel/templates/deployment.yaml
+++ b/security/citadel/templates/deployment.yaml
@@ -38,7 +38,7 @@ spec:
             {{- if .Values.global.sds.enabled }}
             - --sds-enabled=true
             {{- end }}
-            - --append-dns-names=true
+            - --append-dns-names={{ .Values.security.appendDnsNames }}
             - --grpc-port=8060
             - --citadel-storage-namespace={{ .Release.Namespace }}
           {{- if .Values.kustomize }}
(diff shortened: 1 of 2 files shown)


costinm:
Trying to reduce the number of options exposed to users: what is the core feature that the user should configure ? 

If this is about user installing CertManager to generate certs for webhooks - we need to be clear about it, and expose that as an API. Citadel may read the user-friendly config, or just keep generating DNS SANs that will not be used. 

We need to stop thinking of the config as a developer-oriented hashamp and consider the users/operators are the main consumers.

lei-tang:
@costinm as pointed out in the description of this PR,  this PR is related to https://github.com/istio/istio/issues/8736. 
The --append-dns-names flag of Citadel determines whether Citadel generates webhook certificates for Galley and Sidecar Injector. When Istio Webhook Controller is used to generate webhook certificates, a user can set the flag --append-dns-names of Citadel as false so Citadel no longer generates webhook certificates, which simplfies and reduces the work Citadel. 

Can you provide the reasons that you close this PR?

costinm:
It is not a setting that we need to add - in the new installer we are trying to reduce the number of options and simplify the install. 

Right now there is no webhook controler, so the option doesn't make sense. 

It is possible for users to use CertManager and other tools to customize how the certs for webhooks are generated - but because we need to mount both certificates it is not sufficient to ask citadel to not generate.

In other words - it'll not work. Try to write a test - and you'll see that injector or galley will not start if the option is set. 

/hold

costinm:
In other words: happy to approve if you add a test that shows how this option can be set and the webhooks will still work. However we'll still need to consider the usability - what is the user intent, how do we detect and configure this extra component ( CertManager and few other tools can be used ).

Do we want user to configure citadel to not generate, or do we want a higher level config ?

lei-tang:
The option enableGenerateWebhookCerts has a default value of true, which is the same as the current setting of --append-dns-names. That is to say, with its default value being true, the option enableGenerateWebhookCerts will not affect the working of webhooks. After Istio Webhook Controller is deployed, a user can set the option enableGenerateWebhookCerts as false to disable Citadel generating webhook certificates. 

Comparing this option enableGenerateWebhookCerts versus a higher level option, this option provides a more granular control (i.e., control the Citadel) and offers more flexibility.

rlenglet:

PR #313, opened 2019-08-08
Add missing pieces to the demo, and cleanup.

costinm:
Tried to group a bit the files, and moved all alloc configuration to a separate patch file, as recommended by kustomize.

This adds back policy and kiali which were missing in the demo.


istio-policy/templates/config.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if not (eq .Release.Namespace "istio-system") }}
```
howardjohn: This means you can't install policy without telemetry?

howardjohn: I get why you did it, I ran into the same problem. Not sure a better way around it

costinm: Yes, the configs are duplicated in telemetry - I believe @douglas-reid mentioned they are not needed in both places, but I don't know what is used where.

howardjohn: Yeah but now you cant have just policy and no telemetry installed, or this will be skipped?

costinm: Good point - but I don't think we need to worry so much about it with mixer v2 coming.
If @douglas-reid  can identify which configs are needed only for policy we can fix it, but I don't think it's a blocker. For 1.3 we only support demo profile ( all installed ) and we have the default profile that has telemetry but no policy. We don't support policy only in 1.3

howardjohn: Sounds good to me

douglas-reid: There is nothing in this file that is needed *only* for policy except for the `istio-policy` DestinationRule. The rest of this config (attributemanifest, etc.) is shared by all instances of Mixer.

costinm: The problem is 'a-la-carte': some people may install only policy, others only telemetry, and others both. If they install both - we end up with same resource from 2 different components. 
Helm and kustomize don't like this. 

I guess we can add yet another option - but given how hard it is to remove options I think it's 
reasonable to not support "policy without telemetry in istio-system" for 1.3
It is still possible to use operator or installer to install policy in a different namespace, so it doesn't interfere with telemetry.

PR #314, opened 2019-08-09
Fix telemetry labels

howardjohn:
The service was not actually selecting the pod.

Most of the changes are just generated stuff


istio-telemetry/mixer-telemetry/templates/service.yaml
```diff
@@ -18,7 +18,8 @@ spec:
   - name: prometheus
     port: 42422
   selector:
-    app: istio-telemetry
+    istio: mixer
```
canthefason: after spending some time with debugging why mixer is not getting any metrics, I got into the same conclusion as the service/pod labels don't match. While I was planning to push the change, I noticed this PR already exists. 

At least, I can confirm that the label changes you have here works as expected. 👍

PR #315, opened 2019-08-09
Fix prom operator variable

howardjohn:


huang195:
I believe this has been fixed here: https://github.com/istio/installer/pull/311/files

howardjohn:
I don't really understand because my change already has your change in the git tree.. maybe I am missing something obvious

huang195:
oops there were 2 places this variable was added, the PR only fixed 1.

PR #316, opened 2019-08-09
Sync changes to old installer

howardjohn:
Cherrypicks #13392, #13666, #13982, #14077, #14059

### DIFF
--- a/istio-telemetry/kiali/templates/clusterrolebinding.yaml
+++ b/istio-telemetry/kiali/templates/clusterrolebinding.yaml
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: istio-kiali-admin-role-binding-{{ .Release.Namespace }}
+  name: kiali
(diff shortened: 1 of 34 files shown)


howardjohn:
@costinm I added a ton of changes since you approved, should he all but a few I intentionally skipped since they were you hard for one pr

istio-telemetry/kiali/templates/clusterrolebinding.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: istio-kiali-admin-role-binding-{{ .Release.Namespace }}
+  name: kiali
```
costinm: This would prevent installing kiali in multiple namespaces. But it's probably ok, we want to move all cluster-wide configs to separate and kiali is not critical.

howardjohn: I believe the PR was a backwards incompatibility thing

PR #318, opened 2019-08-09
Add Stackdriver tracing metrics and enable authentication methods

canthefason:
This PR will allow Mixer to send Stackdriver tracing metrics upon request, in addition to allowing operators to opt-in for stackdriver authentication methods.


istio-telemetry/mixer-telemetry/values.yaml
```diff
@@ -53,6 +53,15 @@ mixer:
...
       enabled: false
 
+      auth:
+        appCredentials: false
+        apiKey: ""
+        serviceAccountPath: ""
+
+      tracer:
+        enabled: false
+        sampleProbability: 1
```
howardjohn: We already have a value `traceSampling` I think

istio-telemetry/mixer-telemetry/values.yaml
```diff
@@ -53,6 +53,15 @@ mixer:
     stackdriver:
       enabled: false
 
+      auth:
+        appCredentials: false
+        apiKey: ""
+        serviceAccountPath: ""
+
+      tracer:
```
howardjohn: Tracing is defined in like 5 places already... I have no clue how to configure it. Why is this a new field instead of using the existing values?

costinm: I think this is the right place, since mixer is sending the requests.

The others are in global since they have to be injected ( sidecar is sending the data )

howardjohn: We have:
* enableTracing
* tracer
* tracing.enabled
* proxy.tracer

and now mixer.tracer

Maybe this one is right and the others are wrong, but our tracing config is crazy... worse than logging I think

istio-telemetry/mixer-telemetry/values.yaml
canthefason: I know this is confusing, but the confusion is a bit stemming from the API itself. Pilot has a traceSampling field to set this and its value ranges from 0 to 100. 

For this sampleProbability, the range is between 0-1, and acts like a second level filter after what pilot enforces proxies. That said, I'm ok with removing this field completely and set the value to 1 by default.

istio-telemetry/mixer-telemetry/values.yaml
canthefason: When I decided to enable tracing, I also got confused about all those config options. For a while, I thought standing up zipkin was required for enabling tracing since all proxies are set up to send traces to zipkin.istio-system by default and assumed that zipkin will be responsible for sending traces to Stackdriver -which definitely was not true. 

Maybe adding sidecar to stackdriver support might make this less confusing opposed to mixer sending the data to SD.

PR #322, opened 2019-08-13
Add als tls and keepalive config

hanahmily:
as indicated in the title.

Signed-off-by: Gao Hongtao <hanahmily@gmail.com>

### DIFF
--- a/gateways/istio-ingress/templates/deployment.yaml
+++ b/gateways/istio-ingress/templates/deployment.yaml
@@ -145,8 +145,10 @@ spec:
           - {{ $gateway.envoyMetricsService.host }}:{{ $gateway.envoyMetricsService.port }}
         {{- end }}
         {{- if .Values.global.proxy.envoyAccessLogService.enabled }}
-          - --envoyAccessLogServiceAddress
-          - {{ .Values.global.proxy.envoyAccessLogService.host }}:{{ .Values.global.proxy.envoyAccessLogService.port }}
+          - --envoyAccessLogService
+          {{- with  .Values.global.proxy.envoyAccessLogService }}
+          - '{"address":"{{ .host }}:{{.port }}"{{ if .tlsSettings }},"tlsSettings":{{ .tlsSettings | toJson }}{{- end }}{{ if .tcpKeepalive }},"tcpKeepalive":{{ .tcpKeepalive | toJson }}{{- end }}}'
+          {{- end }}
         {{- end }}
         {{- if $gateway.applicationPorts }}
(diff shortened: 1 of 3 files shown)


hanahmily:
CI fails because the master branch of this repo picks RELEASE branch daily build image to run tests instead of the MASTER branch. I added a new function `structToJSON` is not included in release image.

howardjohn:
Where do you see it is pulling the release branch?

```
$ rg tag:
global.yaml
41:  tag: master-latest-daily
```

hanahmily:
> Where do you see it is pulling the release branch?
> 
> ```
> $ rg tag:
> global.yaml
> 41:  tag: master-latest-daily
> ```

@howardjohn 

Sorry for not to verify it carefully. It's not about pulling the release branch. After I dug into it, found it related to `istioctl` in `istionightly/ci:2019-07-01`. It doesn't contain `structToJSON` function, so when ci run `istioctl kube-inject --injectConfigFile istio-control/istio-autoinject/files/injection-template.yaml `, there's an error indicate the missing of it.

If anyone can push a new `istionightly/ci,` I guess ci errors will disappear.

howardjohn:
ah yeah that is it, I'll try to update today

costinm:
To be clear: this is for master, not 1.3 ?

gateways/istio-ingress/templates/deployment.yaml
```diff
@@ -145,8 +145,10 @@ spec:
           - {{ $gateway.envoyMetricsService.host }}:{{ $gateway.envoyMetricsService.port }}
         {{- end }}
         {{- if .Values.global.proxy.envoyAccessLogService.enabled }}
-          - --envoyAccessLogServiceAddress
-          - {{ .Values.global.proxy.envoyAccessLogService.host }}:{{ .Values.global.proxy.envoyAccessLogService.port }}
+          - --envoyAccessLogService
+          {{- with  .Values.global.proxy.envoyAccessLogService }}
```
costinm: This seems far too much template... 

Can we just change the setting to include the expected string ? If that's the syntax of the flag, let's just let user use whatever envoy documents and not attempt to invent our own syntax ( and associated template complexity )

global.yaml
```diff
@@ -98,6 +98,18 @@ global:
       enabled: false
       host: # example: accesslog-service.istio-system
       port: # example: 15000
+      tlsSettings:
```
costinm: How about replacing this with 'envoyAccessLogService' - and a link to wherever this is documented for the structure.

PR #323, opened 2019-08-13
Telemetry is broken when controlPlaneSecurity is disabled

canthefason:
#299 introduced another regression, and now telemetry metrics are no longer received for insecure control planes. 

I'm not sure if this is the right way to fix this issue, but at least it's an attempt parallel to what was intended with #299.

In addition to the fix, this PR also copies the istio-telemetry changes into istio-policy to make them consistent.

### DIFF
--- a/istio-policy/templates/deployment.yaml
+++ b/istio-policy/templates/deployment.yaml
@@ -75,12 +75,17 @@ spec:
 {{- end }}
         imagePullPolicy: {{ .Values.global.imagePullPolicy | default "Always" }}
         ports:
+        - containerPort: 9091
         - containerPort: 15014
         - containerPort: 42422
         args:
           - --monitoringPort=15014
           - --address
+{{- if .Values.global.controlPlaneSecurityEnabled }}
           - unix:///sock/mixer.socket
+{{- else }}
+          - tcp://0.0.0.0:9091
+{{- end }}
 {{- if .Values.global.logging.level }}
           - --log_output_level={{ .Values.global.logging.level }}
 {{- end}}
@@ -138,6 +143,7 @@ spec:
             port: 15014
           initialDelaySeconds: 5
           periodSeconds: 5
+{{- if .Values.global.controlPlaneSecurityEnabled }}
       - name: istio-proxy
 {{- if contains "/" .Values.global.proxy.image }}
         image: "{{ .Values.global.proxy.image }}"
@@ -146,7 +152,6 @@ spec:
 {{- end }}
         imagePullPolicy: {{ .Values.global.imagePullPolicy | default "Always" }}
         ports:
-        - containerPort: 9091
         - containerPort: 15004
         - containerPort: 15090
           protocol: TCP
@@ -209,5 +214,6 @@ spec:
         {{- end }}
         - name: uds-socket
           mountPath: /sock
+{{- end }}
 ---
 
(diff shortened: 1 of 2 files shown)


huang195:
I was following what `istio-control/istio-discovery` was doing with `.Values.global.controlPlaneSecurityEnabled`. When it is set to false, it disables the sidecar from getting injected (https://github.com/istio/installer/blob/master/istio-control/istio-discovery/templates/deployment.yaml#L144).

But I do see the point that disabling the use of TLS in control plane traffic should not prevent the collection of telemetry data. At the moment, they seem to be correlated.

canthefason:
friendly ping since telemetry is broken when `controlPlaneSecurityEnabled=false` which makes this PR critical.

huang195:
+1 as I was able to reproduce the problem
@canthefason when you have `controlPlaneSecurityEnabled=false`, I'm assuming Citadel is disabled, thus not generating certificates. I'm running the same setup, but Envoy side seems to be having issues creating listeners when Citadel is disabled, see below: Wondering if you had run into similar issue.

```
[2019-08-07 19:35:46.233][19][warning][config] [bazel-out/k8-opt/bin/external/envoy/source/common/config/_virtual_includes/grpc_stream_lib/common/config/grpc_stream.h:87] gRPC config stream closed: 14, no healthy upstream
[2019-08-07 19:35:46.233][19][warning][config] [bazel-out/k8-opt/bin/external/envoy/source/common/config/_virtual_includes/grpc_stream_lib/common/config/grpc_stream.h:50] Unable to establish new stream
[2019-08-07 19:35:47.667][19][warning][config] [external/envoy/source/common/config/grpc_mux_subscription_impl.cc:72] gRPC config for type.googleapis.com/envoy.api.v2.Listener rejected: Error adding/updating listener(s) 192.168.1.225_8080: Invalid path: /etc/certs/cert-chain.pem, 192.168.1.225_8079: Invalid path: /etc/certs/cert-chain.pem
```

canthefason:
For my case, Citadel was even enabled. 

When controlPlaneSecurityEnabled=true, Telemetry sidecar proxy is configured to listen to port 9091 and it routes the request to the cluster `inbound_9092`, which ends up with sending the request to the unix socket, `unix://sock/mixer.socket`. See [this](https://github.com/istio/installer/blob/master/istio-telemetry/mixer-telemetry/templates/configmap-envoy.yaml#L47) for the Envoy configurations.

With the recent changes, setting controlPlaneSecurityEnabled=false stops running the sidecar for Telemetry which results with not listening to port 9091 any longer. The result of it is, unhealthy upstream, and telemetry data getting dropped out. That's why I switched listening to a tcp port on telemetry instead of a unix socket for this mode.

huang195:
OK, good to know. FYI, I'm trying a minimalistic Istio deployment with just Pilot and Mixer. Without Citadel generating certificates, it's causing all kinds of problems, even though I've read it in a few places that said Citadel is an optional component. Just wanted to check if others in the community have done something similar.

jaygorrell:
This was also the source of https://github.com/istio/istio/issues/12723 I believe. Also relates to https://github.com/istio/istio.io/issues/3844.

costinm:
If deploying without certificates - you need something similar with the 'minimal' profile, which disables the auth plugin ( that in turns generate cert configs ).

We do test this for pilot+ingress and pilot+sidecar - but we haven't tested this with telemetry. I think the plan was to have this use the 'telemetry-lite' or v2.

PR #330, opened 2019-08-14
fix istio-policy that istio-proxy sidecar depending on global.control…

waret:
…PlaneSecurityEnabled


costinm (approved):

global.yaml
```diff
@@ -19,14 +19,15 @@ global:
   # It is assumed that istio-system is running either 1.0 or an upgraded version of 1.1, but only security components are
   # used (citadel generating the secrets).
   istioNamespace: istio-system
-  configNamespace: istio-system
+
+  configNamespace: istio-control
 
   # Telemetry namespace, including tracing.
-  telemetryNamespace: istio-system
+  telemetryNamespace: istio-telemetry
```
howardjohn: We are not installing in separate namespace anymore

howardjohn (changes requested):

PR #333, opened 2019-08-15
add authorization policy CRD

yangminzhu:
The CRD is also added in istio repo: https://github.com/istio/istio/pull/16187

https://github.com/istio/istio/issues/12394


jmazzitelli:
This has some ramifications for the Kiali Operator (it is adding permissions to the kiali role, which the operator will need to know to do). So if/when we integrate with the kiali operator, this is going to have to be implemented over there as well.

Can someone create a github issue over there so we can track this?

https://github.com/kiali/kiali/issues

howardjohn:
Cherrypicked in https://github.com/istio/installer/pull/344 since it needed some rebasing

PR #335, opened 2019-08-15
Fix Citadel clusterrole and tracing image

howardjohn:
Fix two minor things:
* Citadel needs permissions for ns
* Tracing typo in image name

Everything else in this PR is just junk from run the targets to regen
config


geeknoid:
Force merging since Mergify is unhappy.

Monkeyanator:
@sdake citadel needs to be able to read namespace labels after [this change](https://github.com/istio/istio/pull/15503) (and actually needed to before to check a different ns label)

PR #344, opened 2019-08-19
Fix test failures

howardjohn:


crds/files/crd-14.yaml
```diff
@@ -0,0 +1,24 @@
+kind: CustomResourceDefinition
+apiVersion: apiextensions.k8s.io/v1beta1
+metadata:
+  name: authorizationpolicies.security.istio.io
+  labels:
+    app: istio-pilot
+    istio: security
+    heritage: Tiller
```
costinm: Can we get rid of Tiller labels ? Was needed for backward compat in the crd-11,etc.

Also: are we shipping 14 CRDs in 1.3 ?

howardjohn: We either need to ship the 14 crds for 1.3, or properly handle branching in this repo

costinm: I mean - if it's for 1.4 we shouldn't merge it to the 1.3 release branch - but operator and installer must be able to handle 1.3, so this file missing should not break any test. 

I know master and 1.4 will require it - but want to make sure 1.3 is not going to be affected.

PR #346, opened 2019-08-20
Fix indentation and custom dns names

richardwxn:
fix indentation of egressgateway env
fix custom dns name


howardjohn:
We had some breakages and circle won't automatically rebase. Can you rebase this?

PR #352, opened 2019-08-23
Align prometheus operator job labels with grafana

howardjohn:


howardjohn:
we want to use this for release qualification. Without these changes 99% of the metrics don't show up on the dashboards

PR #359, opened 2019-08-28
Merge tracing provider

clyang82:
Merge tracing provider as one deployment to enable upgrade without use `--force`


clyang82:
> I'm not sure I like the spaghetti deployment - if anything the tracing chart should be split into 3 charts and possibly replaced by the operators for each component.
> 
> The goal is to clean things up and modularize - there is no technical reason to mix 3 third-party components together in a single deployment.

@costinm Thanks for your comment. I am doubt this way as well. It seems it is not clean up and hard to maintain. the goal to have this PR is use the same deployment for different tracing provider to enable upgrade or switch providers easy. I am going to keep the different files with the same deployment. Is that OK? Thanks.

clyang82:
@howardjohn in the istio/istio repo, we have Historical baggage that we cannot break the upgrade case. the istio/installer just alpha feature. correct me if I am wrong. Can I apply the fix to use the same app name here? if not, I am going to close this PR. Thanks.

howardjohn:
We have the opposite problem in this repo - we want things to be modular and independent
If anything these should be split out into different charts, and they don't need to share labels - there is no reason you shouldn't be able to run both

PR #367, opened 2019-08-30
allow long protocol detection timeout for test

yxue:
```    
    # Automatic protocol detection uses a set of heuristics to
    # determine whether the connection is using TLS or not (on the
    # server side), as well as the application protocol being used
    # (e.g., http vs tcp). These heuristics rely on the client sending
    # the first bits of data. For server first protocols like MySQL,
    # MongoDB, etc., Envoy will timeout on the protocol detection after
    # the specified period, defaulting to non mTLS plain TCP
    # traffic. Set this field to tweak the period that Envoy will wait
    # for the client to send the first bits of data. (MUST BE >=1ms)

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -120,7 +120,7 @@ global:
     # the specified period, defaulting to non mTLS plain TCP
     # traffic. Set this field to tweak the period that Envoy will wait
     # for the client to send the first bits of data. (MUST BE >=1ms)
-    protocolDetectionTimeout: 10ms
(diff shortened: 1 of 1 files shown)


yxue:
> do all tests need the long timeout ?

For knative and simple, I verified that we need the long timeout in order to pass the test. For other two, I haven't verified. I think it's safe to extend them all.

yxue:
/cherrypick/release-1.3

PR #371, opened 2019-09-02
Add script to check if newlines are missing

johscheuer:
This is a follow up of https://github.com/istio/istio/pull/16415 which adds missing new lines for existing files. I also added a script that could run these checks automatically I just wanted to ensure that the script is okay before I create an integration with drone CI (which would be an extra step ?)

This time without the additional character :)


morvencao:
Please rebase to pass the tests.

johscheuer:
rebased. Should I add an extra CI step to check for missing newlines ?

morvencao:
Does the linter support newline check?

johscheuer:
It's more a small bash script than a linter. And it evaluates that each file ends with a newline (or more precise a line feed (`\n`) which will also catch the carriage return from Windows (`\r\n`).

morvencao (approved):

PR #378, opened 2019-09-05
Add label name to namespaces

clyang82:
Cannot have sidecar pod running successfully if run this command which is mentioned in https://github.com/istio/installer readme. 
```
    iop istio-control istio-autoinject $IBASE/istio-control/istio-autoinject --set sidecarInjectorWebhook.enableNamespacesByDefault=true --set global.configNamespace=istio-control 
```
That is because the `mutatingwebhookconfiguration` is applied to `istio-sidecar-injector` as well. `  Warning  FailedCreate  0s (x12 over 27s)  replicaset-controller  Error creating: Internal error occurred: failed calling webhook "sidecar-injector.istio.io": Post https://istio-sidecar-injector.istio-control.svc:443/inject?timeout=30s: dial tcp 10.102.56.213:443: connect: connection refused`

the fix is to add label `name` during creating namespace

### DIFF
--- a/bin/iop
+++ b/bin/iop
@@ -55,6 +55,8 @@ function iop() {
         # Apply will make sure that any old config with the same tag that is no longer present will
         # be removed. The release MUST be unique across cluster.
         kubectl create  ns $ns > /dev/null  2>&1
+        #set label 'name' to namespace which is required if enableNamespacesByDefault is true. 
+        kubectl label namespace $ns name=$ns --overwrite
 
         if [ "$ISTIO_ENV" != "" ]; then
             kubectl label namespace $ns istio-env=$ISTIO_ENV --overwrite


bin/iop
```diff
@@ -55,6 +55,8 @@ function iop() {
         # Apply will make sure that any old config with the same tag that is no longer present will
         # be removed. The release MUST be unique across cluster.
         kubectl create  ns $ns > /dev/null  2>&1
+        #set label 'name' to namespace which is required if enableNamespacesByDefault is true. 
+        kubectl label namespace $ns name=$ns --overwrite
```
morvencao: From helm 2.10, helm will add new label `name={{ .Release.Namespace }}` to its created namespaces, see: https://github.com/helm/helm/pull/4231

And our current helm version `2.13.1`, is it an helm issue?

howardjohn: that is only with tiller. we should not rely on tiller

and strongly prefer we don't add a label to the namespace and just fix the webhook config instead -- it's broken in istio/istio for the same reason, there is an open issue

morvencao: @howardjohn it makes sense, we can just remove that matching item from the webhook config.

clyang82: This is original issue to be addressed by adding `name` label - https://github.com/istio/istio/issues/10884 we need to ensure not break the original issue if just remove the `name` matchExpression from webhook config.

clyang82: @howardjohn & @morvencao I cannot find a good solution to handle the open issue in istio/istio. if we just remove `name` label matching from webhook config, it injects the sidecar into `istio-system` which leads sidecar pod cannot be started. any suggestions?

morvencao: The original PR istio/istio#10884 assumes that istio deployment is managed by tiller, and the CI test cases don't cover `sidecarInjectorWebhook.enableNamespacesByDefault=false` case, that's why we haven't found this issue.
IMO it's OK to merge this, wdyt? @howardjohn

howardjohn: we can merge, but this script isn't the supported way to install istio for users, so this will mostly fix the tests not real users? As k believe users will either do helm template or kustomize. I think in kustomize we create the namespace and we set the label, so really only helm template is affected. The same problem exists on istio/istio as far as I know. I agree though, not sure we have a better solution as we can only match on labels.

clyang82: Yes. I found the same problem in istio/istio. cannot find a better solution so far.

morvencao (approved):

PR #387, opened 2019-09-07
add missing pod annotations parameter to istio charts.

morvencao:
merge https://github.com/istio/istio/pull/16886

### DIFF
--- a/istio-cni/templates/daemonset.yaml
+++ b/istio-cni/templates/daemonset.yaml
@@ -27,6 +27,9 @@ spec:
         # priority scheduling and that its resources are reserved
         # if it ever gets evicted.
         scheduler.alpha.kubernetes.io/critical-pod: ''
+        {{- if .Values.cni.podAnnotations }}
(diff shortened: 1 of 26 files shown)


clyang82:
Do we need to have a default `podAnnotations` in global.yaml?

howardjohn:
This is a perfect use case for kustomize or operator overlay. Do we need to
add another API for this?

morvencao:
@howardjohn Istio operator API already support this in k8s overlay, see: https://github.com/istio/operator/blob/master/pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto#L437-L439

howardjohn:
Great to see. If we can support this via kustomize and operator I don't think we should add this to helm values. I believe the goal is to aggressively reduce the amount of config exposed by values.yaml?

morvencao:
@clyang82 not yet, currently many global values, eg. `defaultToleration`, `defaultNodeSelector` haven't been translated to operator API, we may don't want to add another field like that, let's keep it in each single chart.

costinm (approved):

PR #388, opened 2019-09-09
Add back NodePort 31380 31390

clyang82:
To keep consistence with istio/istio. I should be able to access the bookinfo example by using 31380 port.

### DIFF
--- a/gateways/istio-ingress/values.yaml
+++ b/gateways/istio-ingress/values.yaml
@@ -30,8 +30,10 @@ gateways:
     - port: 80
       targetPort: 80
       name: http2
+      nodePort: 31380
     - port: 443
       name: https
+      nodePort: 31390
     - port: 15029
       targetPort: 15029
       name: kiali


howardjohn (changes requested):
having fixed node ports has caused a lot of problems as they need to be globally unique. I don't think we should add these back.

Is there a good reason to add this?

clyang82:
we have defined nodeport in istio/istio with 31380 and 31390 https://github.com/istio/istio/blob/1aeca5f2bd3619072dc522574fc81825ac32d897/install/kubernetes/helm/istio/charts/gateways/values.yaml#L66 and the document claims the default value as well -https://istio.io/docs/reference/config/installation-options/
but it does not work in istio/installer. I think we should keep the same behaviour. WDYT?

howardjohn:
The goal of istio/installer is not to replicate all of the functionality of the old one, but to improve on it. In my opinion adding it in the other installer was a mistake that I don't think we should replicate here, unless there is a good reason to

clyang82:
OK. Thanks for your comments. Let me close it until there is a good reason.

PR #395, opened 2019-09-18
Make use of apps/v1 instead of deprecated extensions

johscheuer:
Small update of the deprecated `extensions/v1beta1` API groups


istio-telemetry/prometheus/templates/inrgess.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.prometheus.ingress.enabled -}}
```
clyang82: Can you help to correct this file name from `inrgess`-->`ingress` along with your changes? Thanks

johscheuer: Oh yes didn't noticed this one.

clyang82:
According to this document - https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/
It seems we have many places need to update. but for the  networking.k8s.io/v1beta1 API, serving Ingress since v1.14. Istio 1.3 has been tested with these Kubernetes releases: 1.13, 1.14, 1.15. the changes will break it.

howardjohn (commented):
CNI change is great, I don't think we should change ingress yet:

```
The v1.20 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:

Ingress (in the extensions/v1beta1 API group)
Migrate to use the networking.k8s.io/v1beta1 API, serving Ingress since v1.14. Existing persisted data can be retrieved/updated via the networking.k8s.io/v1beta1 API.
```

So we have 4 more releases on this versions, and this would break 1.13 which we support

johscheuer:
I just check in helm is an option to check the Kubernetes version `{{ .Capabilities.KubeVersion.Major }}` but this doesn't work with `helm template`. I will reset the Ingress API version.

clyang82 (approved):

morvencao:
@johscheuer The build-in object `.Capabilities` needs tiller support.

PR #396, opened 2019-09-18
Add the possibility to create RoleBinding for PSP

johscheuer:
This allows a user to specify a `psp_cluster_role` which points to a `ClusterRole` that sets the according `PodSecurityPolicy`. With this PR you don't need additional scripts/templates to create the `RoleBinding`.

For the other components some PR's will follow (and docs).

### DIFF
--- a/istio-cni/templates/clusterrolebinding.yaml
+++ b/istio-cni/templates/clusterrolebinding.yaml
@@ -13,3 +13,19 @@ subjects:
 - kind: ServiceAccount
   name: istio-cni
   namespace: {{ .Release.Namespace }}
+---
+{{- if .Values.psp_cluster_role }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: istio-cni-psp
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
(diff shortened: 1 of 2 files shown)


istio-cni/templates/clusterrolebinding.yaml
```diff
@@ -13,3 +13,19 @@ subjects:
 - kind: ServiceAccount
   name: istio-cni
   namespace: {{ .Release.Namespace }}
+---
+{{- if hasKey .Values "psp_cluster_role" }}
```
morvencao: `s/.Values/.Values.cni./`

johscheuer: Thanks, good catch!

istio-cni/templates/clusterrolebinding.yaml
```diff
@@ -13,3 +13,19 @@ subjects:
 - kind: ServiceAccount
   name: istio-cni
   namespace: {{ .Release.Namespace }}
+---
+{{- if hasKey .Values.cni "psp_cluster_role" }}
```
costinm: Not sure what this is, how it'll be used, what the cluster role will be, etc.

Please add a wiki page or some inline docs.

johscheuer: Added some docs in the `values.yaml`

howardjohn (approved):

PR #399, opened 2019-09-20
Make tracing.service.type work

FengyunPan2:
Specify service.type for tracing service, not zipkin service.

Related to: https://github.com/istio/istio/issues/17204

### DIFF
--- a/istio-telemetry/tracing/templates/service.yaml
+++ b/istio-telemetry/tracing/templates/service.yaml
@@ -7,10 +7,9 @@ metadata:
     app: {{ .Values.tracing.provider }}
     release: {{ .Release.Name }}
 spec:
-  type: {{ .Values.tracing.service.type }}
   ports:
     - port: {{ .Values.tracing.service.externalPort }}
-      targetPort: 9411
+      targetPort: {{ .Values.tracing.zipkin.queryPort }}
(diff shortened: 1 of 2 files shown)


istio-telemetry/tracing/templates/service.yaml
```diff
@@ -30,14 +29,15 @@ metadata:
     app: {{ .Values.tracing.provider }}
     release: {{ .Release.Name }}
 spec:
+  type: {{ .Values.tracing.service.type }}
   ports:
-    - name: http-query
-      port: 80
+    - name: {{ .Values.tracing.service.name }}
+      port: {{ .Values.service.externalPort }}
```
morvencao: s/.Values.service.externalPort/.Values.tracing.service.externalPort/

morvencao: not sure if this will have impact for current user, because we're changing default port from `80` to `9411`. otherwise good to me.

FengyunPan2: It's ok, I have tested it.

FengyunPan2:
Any thought?

howardjohn (approved):

FengyunPan2:
@morvencao Thx.

PR #403, opened 2019-09-22
injector: update datadog config in template

cgilmour:
This change is to match istio/istio#16594


morvencao:
Can you rebase? we have removed `integration-presubmit-isolated-namespaces-kind` test.

morvencao (approved):
Can you rebase? we have removed integration-presubmit-isolated-namespaces-kind test.

PR #407, opened 2019-09-26
update hub and tag examples in readme

linsun:
with example of specifying your own hub and tag

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,9 +152,10 @@ For upgrades from the official installer, it is recommended to install the secur
 at the end uninstall the official installer, and lock down istio-system.
 
 This is currently required if any mTLS is used. In future other Spifee implementations can be used, and
-it is possible to use other tools that create the expected certificates for Istio.
+it is possible to use other tools that create the expected certificates for Istio.  The following example shows how to install istio 1.3.0's citadel:
 
 ```bash
+export HUB=docker.io/istio TAG=1.3.0 
 iop istio-system citadel $IBASE/security/citadel
 ```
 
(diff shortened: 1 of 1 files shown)


clyang82:
the test is using the default tag and hub. I think it works. what is your case?

linsun:
@clyang82 i didn't update the test, not sure why it is failing. I was just going through the readme to make it clear.

linsun:
@clyang82 I had issues with corporate firewall so my machine can't access GCR at work (thanks to @sdake) so I had to use released images from docker hub for now before that is resolved. I've updated the PR to make that clear.

clyang82 (approved):

clyang82:
rebase should help.

README.md
```diff
@@ -155,6 +155,14 @@ This is currently required if any mTLS is used. In future other Spifee implement
...
 
 ```bash
+# use the default HUB=gcr.io/istio-release, TAG=master-latest-daily defined in global.yaml
+iop istio-system citadel $IBASE/security/citadel
+```
+
+If you want to use a different `HUB` and `TAG`, the following example shows how to install istio 1.3.0's citadel:
+
+```bash
+export HUB=docker.io/istio TAG=1.3.0
```
sdake: this looks fine - although inconsistent with our developer wiki.  In our developer wiki, we recommend setting HUB and TAG and ISTIO.  See: https://github.com/istio/istio/wiki/Preparing-for-Development#setting-up-environment-variables

I guess I don't know what $IBASE is :)

sdake (approved):

README.md
linsun: If I set per developer wiki, it won't work in this scenario.  I believe TAG istio is wrong and confusing also.

PR #421, opened 2019-10-05
Support DNS certificate provision and management

lei-tang:
This PR is for issue: https://github.com/istio/istio/issues/17061
This PR depends on: https://github.com/istio/istio/pull/17400

Design documents:

### DIFF
--- a/istio-control/istio-discovery/templates/clusterrole.yaml
+++ b/istio-control/istio-discovery/templates/clusterrole.yaml
@@ -34,5 +34,16 @@ rules:
 - apiGroups: [""]
   resources: ["endpoints", "pods", "services", "namespaces", "nodes", "secrets"]
   verbs: ["get", "list", "watch"]
+{{ if .Values.pilot.enableCertificateController }}
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["create", "get", "watch", "list", "update", "delete"]
+- apiGroups: ["certificates.k8s.io"]
+  resources:
+    - "certificatesigningrequests"
+    - "certificatesigningrequests/approval"
+    - "certificatesigningrequests/status"
+  verbs: ["update", "create", "get", "delete"]
+{{ end }}
 ---
 {{ end }}
--- a/istio-control/istio-discovery/templates/configmap.yaml
+++ b/istio-control/istio-discovery/templates/configmap.yaml
@@ -211,5 +211,13 @@ data:
       envoyAccessLogService:
         address: {{ .Values.global.proxy.envoyAccessLogService.host }}:{{ .Values.global.proxy.envoyAccessLogService.port }}
     {{- end}}
+
+    {{- if .Values.pilot.enableCertificateController }}
+      certificates:
+        - secretName: istio-webhook-galley
+          dnsNames: [istio-galley.{{ .Release.Namespace }}.svc, istio-galley.{{ .Release.Namespace }}]
+        - secretName: istio-webhook-sidecar-injector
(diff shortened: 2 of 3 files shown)


howardjohn (approved):

istio-control/istio-discovery/templates/configmap.yaml
```diff
@@ -211,5 +211,13 @@ data:
       envoyAccessLogService:
         address: {{ .Values.global.proxy.envoyAccessLogService.host }}:{{ .Values.global.proxy.envoyAccessLogService.port }}
     {{- end}}
+
+    {{- if .Values.pilot.enableCertificateController }}
```
costinm: Same as above: a 'certificates' section under values.global, no need for an 'enable' API.

istio-control/istio-discovery/templates/configmap.yaml
```diff
@@ -211,5 +211,13 @@ data:
...
+      certificates:
+        - secretName: istio-webhook-galley
```
costinm: Let's not use random names for the secretName. 
Istio spiffe secrets follow the pattern: istio.SERVICE_ACCOUNT_NAME

For the DNS certs we should follow a similar convention - istiodns or dns prefix ( I would use dns. to make it clear it's not istio specific - unlike the Spifee accounts that are specific to Istio due the SAN format ). And the second part should be based on either the service account or service name. 

While this won't be enforced, it's good to be organized and have a good style.

istio-control/istio-discovery/values.yaml
```diff
@@ -100,6 +100,10 @@ pilot:
   # Indicate if Galley is enabled to send MCP queries
   useMCP: true
 
+  # When true, DNS certificates will be provisioned through Chiron linked into Pilot.
+  # When false, such mechanism is not enabled.
+  enableCertificateController: false
```
costinm: Move this to top level global, and rename to 'certificates:' 

In mesh config - add the content of this section using 'toJSON' - like we do for networks.

costinm: And we want this enabled by default - so it gets tested. We can turn it off before the release. ( like we did with sniffing)

costinm: I mean: enabled by default now, we'll decide to enable it or not before the release. We need to see what happens on K8S clusters that don't support the API.

istio-control/istio-discovery/templates/clusterrole.yaml
```diff
@@ -34,5 +34,16 @@ rules:
 - apiGroups: [""]
   resources: ["endpoints", "pods", "services", "namespaces", "nodes", "secrets"]
   verbs: ["get", "list", "watch"]
+{{ if .Values.pilot.enableCertificateController }}
```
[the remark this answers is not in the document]
lei-tang: Done: addressed in https://github.com/istio/installer/pull/424; please review it.

istio-control/istio-discovery/templates/configmap.yaml
lei-tang: Done: addressed in https://github.com/istio/installer/pull/424; please review it.

istio-control/istio-discovery/templates/configmap.yaml
lei-tang: Done: addressed in https://github.com/istio/installer/pull/424; please review it.

istio-control/istio-discovery/values.yaml
lei-tang: Done: addressed in https://github.com/istio/installer/pull/424; please review it.

PR #422, opened 2019-10-05
Support Citadel deployment to add random env variables

JimmyCYJ:
Per review comment on https://github.com/istio/istio/pull/17550#discussion_r331690392, add the same deployment change into installer for Citadel.

https://github.com/istio/istio/issues/17059

### DIFF
--- a/security/citadel/templates/deployment.yaml
+++ b/security/citadel/templates/deployment.yaml
@@ -70,6 +70,12 @@ spec:
           env:
           - name: CITADEL_ENABLE_NAMESPACES_BY_DEFAULT
             value: "{{ .Values.security.enableNamespacesByDefault }}"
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

security/citadel/templates/deployment.yaml
```diff
@@ -70,6 +70,12 @@ spec:
           env:
           - name: CITADEL_ENABLE_NAMESPACES_BY_DEFAULT
             value: "{{ .Values.security.enableNamespacesByDefault }}"
+          {{- if .Values.env }}
```
morvencao: SHouldn't it be `.Values.security.env`, have raised a followup PR to fix this: https://github.com/istio/installer/pull/423

PR #423, opened 2019-10-05
fix values path.

morvencao:
fix incorrect values path of https://github.com/istio/installer/pull/422


howardjohn (approved):
thanks! I'm working on getting the linters running so we can catch all of these - there is a few more I've found

PR #424, opened 2019-10-07
Enable DNS certificate provisioning

lei-tang:
This PR is for issue: istio/istio#17061

Design documents:

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -464,6 +464,15 @@ global:
   # will ignore the helm test yaml files when generating the template
   enableHelmTest: false
 
+  # Configures DNS certificates provisioned through Chiron linked into Pilot.
+  # Te values in this file are all hard-coded; please ensure the namespaces
+  # in dnsNames are consistent with those of your services.
+  certificates:
+    - secretName: dns.istio-galley-service-account
+      dnsNames: [istio-galley.istio-system.svc, istio-galley.istio-system]
+    - secretName: dns.istio-sidecar-injector-service-account
+      dnsNames: [istio-sidecar-injector.istio-system.svc, istio-sidecar-injector.istio-system]
+
 # Internal setting - used when generating helm templates for kustomize.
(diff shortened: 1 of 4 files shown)


global.yaml
```diff
@@ -464,6 +464,15 @@ global:
   # will ignore the helm test yaml files when generating the template
   enableHelmTest: false
 
+  # Configures DNS certificates provisioned through Chiron linked into Pilot.
+  # Te values in this file are all hard-coded; please ensure the namespaces
+  # in dnsNames are consistent with those of your services.
+  certificates:
+    - secretName: dns.istio-galley-service-account
+      dnsNames: [istio-galley.istio-system.svc, istio-galley.istio-system]
```
rlenglet: Any way to parameterize these DNS names with the `istioNamespace` setting?

global.yaml
```diff
@@ -464,6 +464,15 @@ global:
   # will ignore the helm test yaml files when generating the template
   enableHelmTest: false
 
+  # Configures DNS certificates provisioned through Chiron linked into Pilot.
+  # Te values in this file are all hard-coded; please ensure the namespaces
```
rlenglet:
```suggestion
  # The DNS names are all hard-coded; please ensure the namespaces
```
Point to which values need to be kept in sync.
And in the comments for those values, add comments to point back to here.

global.yaml
lei-tang: I have the same question; all the values in this file are hard-coded.
Add @costinm and @howardjohn regarding to this question.

global.yaml
lei-tang: Done. For the config in istio-control/istio-discovery/templates/configmap.yaml that uses this field, the config already points back to here as: .Values.global.certificates.

global.yaml
howardjohn: Not that I know of unfortunately

howardjohn (approved):

PR #425, opened 2019-10-07
use global.network to set ISTIO_META_NETWORK for sidecars and ingressgateway

ayj:
Reduce how many duplicate fields an operator needs to set for networking. As far as I can tell `global.network` and `gateways.istio-ingressgateway.env.ISTIO_META_NETWORK` typically have the same value. The former sets the ISTIO_META_NETWORK env variable for the injected sidecars. The latter sets the env variable for the ingressgateway. This PR uses `global.network` to set both env variables if the istio-ingressgateway was not explicitly configured.


ayj:
cc @costinm

gateways/istio-ingress/templates/deployment.yaml
```diff
@@ -245,6 +245,10 @@ spec:
             value: {{ $val }}
           {{- end }}
           {{- end }}
+          {{- if and (not $gateway.env.ISTIO_META_NETWORK) .Values.global.network }}
+          - name: ISTIO_META_NETWORK
+            value: {{ .Values.global.network }}
+          {{- end }}
```
ayj: @rshriram, is there ever any reason the value of ISTIO_META_NETWORK should differ between the sidecar and ingress-gateway proxies in a mesh?

rshriram: no.. the only usecase is the egress gateway

ayj: This PR should be good then if they never differ. One less option to configure.

I don't see ISTIO_META_NETWORK used for egress but I do see ISTIO_META_REQUESTED_NETWORK_VIEW.

howardjohn (approved):

PR #428, opened 2019-10-08
Configure webhooks of Galley and Sidecar Injector based on DNS certificate provisioning

lei-tang:
Configure the webhook configurations of Galley and Sidecar Injector based on DNS certificate provisioning: (the approach is this PR is same as that of https://github.com/istio/istio/pull/16118 and https://github.com/istio/istio/pull/16121).
- If DNS certificates are provisioned, Galley and Sidecar Injector will not configure their own webhook configurations.
- Otherwise, Galley and Sidecar Injector configure their own webhook configurations (their current behavior).

This PR depends on: https://github.com/istio/istio/pull/16121

Please provide a description for what this PR is for: https://github.com/istio/istio/issues/17061

Design documents:

### DIFF
--- a/global.yaml
+++ b/global.yaml
@@ -479,11 +479,17 @@ global:
   # Configures DNS certificates provisioned through Chiron linked into Pilot.
   # The DNS names in this file are all hard-coded; please ensure the namespaces
   # in dnsNames are consistent with those of your services.
-  certificates:
-    - secretName: dns.istio-galley-service-account
-      dnsNames: [istio-galley.istio-system.svc, istio-galley.istio-system]
-    - secretName: dns.istio-sidecar-injector-service-account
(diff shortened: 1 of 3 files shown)


howardjohn (approved):

costinm (commented):
Looks good - but I think in the short term we need another (temp) setting since istioctl and reconcile webhook is not yet ready ? 

Maybe "Values.global.operator"  - to be set when operator is used to install istio ( and as such takes care of webhooks ) - and false if helm / helm template are used ?

costinm:
To be clear: I think we should have certificates enabled by default, so it gets tested. We may turn it off before release. 

But reconcile webhook will still be needed for non-operator cases and default install.

lei-tang:
@costinm The PR has been revised based on your comments. Please take a look. Thanks!

PR #430, opened 2019-10-09
Use dev build instead of daily builds

howardjohn:
Daily builds are no longer published

Also we don't have gencheck on this repo so some random changes are
included that are generated.


kustomize/default/gen-istio-discovery.yaml
```diff
@@ -254,7 +255,7 @@ data:
     enableTracing: true
 
     # Set accessLogFile to empty string to disable access log.
```
sdake: may want to change the comment.

kustomize/default/gen-istio-discovery.yaml
```diff
@@ -338,6 +339,20 @@ data:
       #
       # Address where istio Pilot service is running
       discoveryAddress: istio-pilot.istio-system:15011
+
+      # Configures DNS certificates provisioned through Chiron linked into Pilot.
```
sdake: I thought we were not mentioning Chiron anywhere?

kustomize/istio-canary/gen-discovery.yaml
```diff
@@ -255,7 +256,7 @@ data:
     enableTracing: true
 
     # Set accessLogFile to empty string to disable access log.
```
sdake: comment doesn't match code

kustomize/istio-canary/gen-discovery.yaml
```diff
@@ -337,6 +338,20 @@ data:
       #
       # Address where istio Pilot service is running
       discoveryAddress: istio-pilot.istio-system:15011
+
+      # Configures DNS certificates provisioned through Chiron linked into Pilot.
```
sdake: chiron

kustomize/minimal/gen-discovery.yaml
```diff
@@ -255,7 +255,7 @@ data:
     enableTracing: true
 
     # Set accessLogFile to empty string to disable access log.
```
sdake: comment on access log

sdake (commented):
The choices for ImagePullPolicy don't seem quite right.  There are a few components that have hard tags (rather than `:latest`.  No good reason to force K8s to repull these).  I'd have concerns about using `:latest` as well as docker pull pilot will pull latest first.  This may not be desireable.

I'll mark this as a hold, then approve.  If you want to make changes in this PR, thats cool, if not simply remove the hold.

Cheers
-steve

sdake (approved):
as per prior review, approving (remove the hold to merge).

PR #437, opened 2019-10-16
Update for release-1.4

howardjohn:


test/knative/values.yaml
```diff
@@ -54,5 +54,5 @@ global:
 
 
   proxy_init:
-    image: proxy_init
+    image: proxyv2
```
jwendell: This should be done in master as well

jwendell (approved):

test/knative/values.yaml
howardjohn: fixed in https://github.com/istio/installer/pull/438

PR #441, opened 2019-10-18
When Operator manages webhooks, Galley and Sidecar Injector are deprivileged

lei-tang:
Please provide a description for what this PR is for: https://github.com/istio/istio/issues/17061

Design documents:

[ ] Docs
[] Installation
[X ] Networking
[ ] Policies and Telemetry
[X ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

### DIFF
--- a/istio-control/istio-autoinject/templates/clusterrole.yaml
+++ b/istio-control/istio-autoinject/templates/clusterrole.yaml
@@ -11,7 +11,9 @@ rules:
   resources: ["configmaps"]
   resourceNames: ["istio-sidecar-injector"]
   verbs: ["get", "list", "watch"]
+{{- if not .Values.global.operatorManageWebhooks }}
 - apiGroups: ["admissionregistration.k8s.io"]
   resources: ["mutatingwebhookconfigurations"]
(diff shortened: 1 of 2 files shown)


ayj (commented):
Did you test the re-configuration case of moving from operatorManageWebhooks=false to operatorManageWebhooks=true? Things should eventually converge if the operator is started successfully. I'm not sure how much disruption this will cause during the switch. Best case I think we end up spamming the logs with permission denied for watch/list.

lei-tang:
operatorManageWebhooks is a helm value used in installation. If operatorManageWebhooks is changed from false to true, in the Istio installed, webhooks no longer manage their webhooks. API server either 1) not send any webhook requests before the operator command is executed to create webhook configurations or 2) send webhook requests to the original webhook configurations if such configurations exist. For the case 2), the duration of disruption during switching to the operator configurations depends on how soon k8s API server updates webhook configurations, which should be fast but I have not measured it yet.

howardjohn (approved):

PR #450, opened 2019-10-22
create MeshPolicy using global.mtls.enabled

ayj:
The helm charts from istio/istio create MeshPolicy when
global.mtls.enabled=true. The new istio/installer charts and operator
do not despite the API defining a global.mtls.enabled option. Long
term the operator should update MeshPolicy based on the MeshConfig. In
the short term (~1.4) the istio/installer and operator will be updated
to match the existing behavior to not surprise user

ref: https://github.com/istio/istio/issues/18199


ayj:
cc @ostromart

howardjohn (approved):

ayj:
cherrypick/release-1.4

PR #458, opened 2019-10-24
Fix kind cleanup to prevent leak

howardjohn:
This is causing a resource leak, causing test failures across the org


howardjohn:
https://github.com/istio/test-infra/issues/1988

clarketm (approved):

costinm (approved):

PR #457, opened 2019-10-24
Generate files and add gencheck target

howardjohn:

PR #462, opened 2019-10-25
cleanup the make run-build-demo

incfly:
ideally should have a ci/cd with  presubmit check...otherwise out of sync issue for test/demo vs. root config.

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -239,6 +239,9 @@ info:
 	env
 	id
 
+gen: run-build
+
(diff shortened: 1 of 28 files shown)


incfly:
@howardjohn finally understand why `k apply -k github.com/../deom/` not working... there're some gen file template in the demo folder needs to be manually updated as `make`

incfly:
This one does not have anything other than run make against head of master.

howardjohn:
See https://github.com/istio/installer/pull/457, which enforces this as CI

PR #465, opened 2019-10-26
remove the usage of applicationPorts

yangminzhu:
This should fix the test fail of https://github.com/istio/istio/issues/18351, the `applicationPort` is removed in https://github.com/istio/istio/pull/17986.


howardjohn (approved):
Thanks!

PR #467, opened 2019-10-27
Add mixerv2 telemetry install option

richardwxn:


howardjohn (commented):
Shouldn't we just be adding a new "mixerv2" option, and keeping the old option the same? If they enable mixerv2, then install the Envoyfilter and disable http for mixerv1.

Also I don't think `mixerv2` is the name we are going with, although I am not sure what the correct name is

istio-control/istio-discovery/values.yaml
```diff
@@ -96,8 +96,14 @@ pilot:
     ingressClass: istio
 
   telemetry:
-    # Will not define mixerCheckServer and mixerReportServer
-    enabled: true
+    mixerv1:
```
mandarjog: kill the name Mixer. We should not add anything that says `mixer` anywhere :-)

1. `telemetry.enabled: false` should mean turn off all telemetry.
so please restore that option.

2. rename options to `v1` and `v2` 
Fix comment.
v1.enabled `will` define mixerCheck and report servers.

richardwxn: what if one specifies telemetry.enabled: true and v1.enabled: true and v2.enabled: true? Which should actually take effect? We can add corresponding validation logic in operator but not sure how to deal with the values.yaml case.

istio-control/istio-discovery/templates/telemetryv2.yaml
```diff
@@ -0,0 +1,113 @@
+{{- if and .Values.pilot.telemetry.enabled .Values.pilot.telemetry.v2.enabled }}
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
```
mandarjog: The authoritative source of this configuration will be in istio/istio e2e tests.
Can you restructure it so that the template "reads" from a local file. We can easily keep the local file in sync with the istio/istio

richardwxn: I added two separate local files for metadata-exchange filter and stats filter

mandarjog (approved):
/lgtm

richardwxn:
@howardjohn Can you PTAL?

howardjohn (approved):

PR #469, opened 2019-10-29
make run-build after setting global.mtls.auto true in globa.yaml

incfly:
https://github.com/istio/istio/issues/14524


kustomize/citadel/citadel.yaml
```diff
@@ -1,3 +1,22 @@
+---
+# Source: citadel/templates/poddisruptionbudget.yaml
+
+apiVersion: policy/v1beta1
```
howardjohn: weird this changed...

howardjohn (approved):
you must have a different version of helm or something than CI? maybe try `BUILD_WITH_CONTAINER=1 make build` so its the same as CI?

incfly:
`BUILD_WITH_CONTAINER make gen` is the key.

PR #476, opened 2019-11-02
Sync CNI charts from istio/cni 1.4 branch

jwendell:
Closes https://github.com/istio/istio/issues/18576


jwendell:
Will open a PR against master and then cherrypick

PR #477, opened 2019-11-02
Sync CNI charts from istio/cni

### LINKED ISSUE #18576: CNI charts are outdated
Previously (or currently) CNI components can be - and are - encouraged to be installed on `kube-system` namespace:

```
$ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=kube-system | kubectl apply -f -
```
This way users can setup additional priorities for pod scheduling like `system-node-critical`.

With the `istioctl manifest` method, those components are installed in `istio-system` namespace.

Is there an way to install them in a different namespace like it's possible with `helm`?

jwendell:
Closes https://github.com/istio/istio/issues/18576


istio-cni/Chart.yaml
```diff
@@ -1,12 +1,12 @@
 apiVersion: v1
 name: istio-cni
-version: 0.2.0
-appVersion: 0.2.0
-tillerVersion: ">=2.7.2-0"
+version: 1.1.0
```
rlenglet: Those 3 lines should probably be dropped.

istio-cni/templates/istio-cni.yaml
```diff
@@ -27,9 +85,6 @@ spec:
         # priority scheduling and that its resources are reserved
         # if it ever gets evicted.
         scheduler.alpha.kubernetes.io/critical-pod: ''
-        {{- if .Values.cni.podAnnotations }}
```
rlenglet: Removing these 3 lines is a regression. It seems like you've not copied the `master` version of the file from `istio/cni`.

rlenglet: And for some reason the bot failed to create a cherry-pick commit for https://github.com/istio/cni/pull/206. Just triggered it. We'll need to merge https://github.com/istio/cni/pull/215 for 1.4.0.

rlenglet (changes requested):

istio-cni/Chart.yaml
jwendell: done

istio-cni/templates/istio-cni.yaml
jwendell: done, thanks!

rlenglet (approved):

istio-cni/templates/istio-cni.yaml
```diff
@@ -79,3 +140,27 @@ spec:
         - name: cni-net-dir
           hostPath:
             path: {{ default "/etc/cni/net.d" .Values.cni.cniConfDir }}
+---
+
+apiVersion: v1
+kind: ServiceAccount
```
howardjohn: Looks like we lost imagepullsecrets?

istio-cni/templates/_labels.tpl
```diff
@@ -0,0 +1,10 @@
+{{- define "common_labels" }}
```
howardjohn: We are trying to avoid these templates as much as possible.

Also we don't want the extra chart/heritage anymore since this will be used by the operator

istio-cni/values_gke.yaml
```diff
@@ -1,6 +1,6 @@
 cni:
-  hub: docker.io/tiswanso
-  tag: v0.1-dev
+  hub: gcr.io/istio-release
+  tag: master-latest-daily
```
howardjohn:
```suggestion
    tag: latest
```

istio-cni/values_gke.yaml
```diff
@@ -1,6 +1,6 @@
 cni:
-  hub: docker.io/tiswanso
-  tag: v0.1-dev
+  hub: gcr.io/istio-release
```
howardjohn:
```suggestion
  hub: gcr.io/istio-testing
```

We need to apply this to values.yaml as well I think

We can probably just delete values_gke.yaml, to avoid duplication. Its wrong to use it already since its not properly kept in sync

jwendell:
@howardjohn Changed the PR to keep the current directory structure, making it smaller and easier to review. Addressed your comments.

howardjohn (approved):
Thanks!

PR #485, opened 2019-11-05
Update envoy metrics service and access log service config

joeyb:
This PR migrates to the newer flags for the envoy metrics service and access log service config that support setting up the TLS context and keepalive config.


howardjohn:
/ok-to-test

Thanks!

howardjohn (approved):

PR #486, opened 2019-11-05
Allow the user to override CNI namespace

### LINKED ISSUE #18640: CNI daemonset  should be able to deploy in kube-system ns
**Original description of https://github.com/istio/istio/issues/18576 :**

Previously (or currently) CNI components can be - and are - encouraged to be installed on `kube-system` namespace:

```
$ helm template install/kubernetes/helm/istio-cni --name=istio-cni --namespace=kube-system | kubectl apply -f -
```
This way users can setup additional priorities for pod scheduling like `system-node-critical`.

With the `istioctl manifest` method, those components are installed in `istio-system` namespace.

Is there an way to install them in a different namespace like it's possible with `helm`?

**This is a valid bug.**

Some systems (Like OpenShift) refuse to deploy pods with high priority in namespaces other than `kube-system`.

See https://github.com/kubernetes/kubernetes/pull/76310

jwendell:
Closes https://github.com/istio/istio/issues/18640.

### DIFF
--- a/istio-cni/templates/clusterrolebinding.yaml
+++ b/istio-cni/templates/clusterrolebinding.yaml
@@ -12,14 +12,14 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: istio-cni
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.cni.namespace | default .Release.Namespace }}
 ---
 {{- if ne .Values.cni.psp_cluster_role "" }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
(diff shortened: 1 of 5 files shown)


howardjohn (commented):
this doesn't seem right, they should just change the release namespace?

howardjohn:
unless this is a workaround for an operator limitations

howardjohn:
I think each component has a namespace field, so something like cni.namespace=something? Not sure if that works

jwendell:
It seems `cni.components.cni.namespace` works, let me try.

jwendell:
It works, thanks.

(BTW, That's a lot confusing!)

PR #489, opened 2019-11-05
Kill unneeded/broken DestinationRules

howardjohn:
See https://github.com/istio/istio/issues/18654

This is especially bad because the config will be reconciled, so even if
manually fixed will still break


howardjohn:
/retest

bug in istio/istio https://github.com/istio/istio/issues/18665

istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
```diff
@@ -63,19 +63,6 @@ spec:
   - mtls:
       mode: PERMISSIVE
 ---
-apiVersion: networking.istio.io/v1alpha3
```
incfly: Line 8 Destinationrule for API server does not exist in the old installer. I think we only need that if `global.mtls.enabled == true`, put it in the else branch, rather than in the top. Unless I miss anything new installer/operator specific.

howardjohn: Fixed, thanks

incfly (approved):
/lgtm

ostromart (approved):

PR #490, opened 2019-11-05
Add missing remote resources

ostromart:


istio-control/istio-discovery/templates/service.yaml
```diff
@@ -25,3 +25,64 @@ spec:
...
+  - port: 15014
+    name: http-monitoring
+  clusterIP: None
+---
+{{- end }}
+{{- if and .Values.global.remotePolicyAddress .Values.global.createRemoteSvcEndpoints }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: istio-policy
```
howardjohn: this doesn't seem right to exist in the pilot chart.. not sure exactly how these are used though?

crds/templates/endpoints.yaml
```diff
@@ -0,0 +1,63 @@
+{{- if or .Values.global.remotePilotCreateSvcEndpoint .Values.global.createRemoteSvcEndpoints }}
```
howardjohn: seems like this shouldn't be in crd, as its not a crd.

Should we have a new "remote chart" that just has this remote stuff? I am not exactly sure how this is intended to be used so not sure what makes the most sense.

ostromart: This is going to have a ripple effect for operator too - we'd have to create a remote feature and component and stuff it into the API.  That seems like a bit change at the 11th hour.
Alternative: we leave it in here for now and decide how best to handle it in 1.5. It would mean a fairly small change to the docs if we do create a new chart for it.

istio-control/istio-discovery/templates/service.yaml
ostromart: I'll put it with the other resources for now - also see comment below.

istio-control/istio-discovery/templates/configmap.yaml
```diff
@@ -41,27 +81,12 @@ data:
...
-    {{- end }}
-
     {{- if .Values.pilot.telemetry.v2.enabled }}
     disableMixerHttpReports: true
     {{- else }}
     disableMixerHttpReports: false
     {{- end }}
 
-
     {{- if .Values.pilot.policy.enabled }}
```
elfinhe: Replace L240-L254 with the lines in istio/istio templates/configmap.yaml:

```

    {{- $defPilotHostname := printf "istio-pilot.%s" .Release.Namespace }}
    {{- $pilotAddress := .Values.global.remotePilotAddress | default $defPilotHostname }}
    {{- if .Values.global.controlPlaneSecurityEnabled }}
      #
      # Mutual TLS authentication between sidecars and istio control plane.
      controlPlaneAuthPolicy: MUTUAL_TLS
      #
      # Address where istio Pilot service is running
      {{- if or .Values.global.remotePilotCreateSvcEndpoint .Values.global.createRemoteSvcEndpoints }}
      discoveryAddress: {{ $defPilotHostname }}:15011
      {{- else }}
      discoveryAddress: {{ $pilotAddress }}:15011
      {{- end }}
    {{- else }}
      #
      # Mutual TLS authentication between sidecars and istio control plane.
      controlPlaneAuthPolicy: NONE
      #
      # Address where istio Pilot service is running
      {{- if or .Values.global.remotePilotCreateSvcEndpoint .Values.global.createRemoteSvcEndpoints }}
      discoveryAddress: {{ $defPilotHostname }}:15010
      {{- else }}
      discoveryAddress: {{ $pilotAddress }}:15010
      {{- end }}
    {{- end }}
```

elfinhe: + I am not confident we are missing something else.

elfinhe: Just confirmed that I should have all "global.values.remote*" synced, with the one I mentioned above.

ostromart: Updated the configmap.

elfinhe (approved):

howardjohn (approved):

ostromart:
Per my chat with @nmittler and @ayj there are changes in the pipeline that will likely make this obsolete. Let's park this review until those changes land.

elfinhe:
> Per my chat with @nmittler and @ayj there are changes in the pipeline that will likely make this obsolete. Let's park this review until those changes land.

Does this mean that .Values.global.remotePilot is not supported in the new installer charts in 1.4.0?

howardjohn (approved):

ayj (commented):
Can you move the [istio-reader-service-account resources](https://github.com/istio/installer/blob/aa7176d9e23c71b71df5a0a1cc813bc89d651940/istio-control/istio-discovery/templates/clusterrole.yaml#L47-L69) into the base after the directory rename?

crds/templates/endpoints.yaml
ostromart: Per offline conversation, renaming to base, which is consistent with operator.

ostromart:
@ayj, I've made the change, please check that we have the correct install qualifier around it.

PR #493, opened 2019-11-05
Use global trust domain

pitlv2109:
For some reason Citadel is using its own trust domain helm flag from the `security.trustDomain`. Every component should use the global trust domain value. Pilot and other components are already using `.Values.global.trustDomain`.


pitlv2109:
@howardjohn any idea why lint and gen_check are failing? I ran `make gen` and included the generated files to this PR.

howardjohn (approved):

howardjohn:
run `BUILD_WITH_CONTAINER=1 make gen` you probably have a different helm version locally, that will build in the CI container

PR #496, opened 2019-11-06
Move up grafana to latest

linsun:
I did a quick testing of 6.4.3 with 1.4 beta 2, didn't notice any issues.


sdake:
looks fine once you run `make gen`.  You may need to unset HUB and TAG if they are set in your enviornments ot run `make gen.

howardjohn:
This is only modifying the generate files not the source of those files. The source of truth is something like istio-telemetry/grafana/values.yaml

linsun:
@howardjohn @sdake thanks, didn't realize the files were generated. PTAL.

howardjohn (approved):

PR #502, opened 2019-11-06
No ISTIO_MUTUAL DestinationRule if auto mTLS is enabled.

incfly:

Starting from installer part.

### DIFF
--- a/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
+++ b/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
@@ -33,8 +33,9 @@ spec:
     tls:
       mode: ISTIO_MUTUAL
 ---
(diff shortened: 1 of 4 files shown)


howardjohn (approved):

istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
```diff
@@ -33,8 +33,9 @@ spec:
     tls:
       mode: ISTIO_MUTUAL
 ---
-# Destination rule to disable (m)TLS when talking to API server, as API server doesn't have sidecar.
-# Customer should add similar destination rules for other services that dont' have sidecar.
+{{- if not .Values.global.mtls.auto }}
```
incfly: hmm. @diemtvu helped realized this is copied pasted in the wrong block. should be upper in the ISTIO_MUTUAL section...

PR #509, opened 2019-11-06
Move auto mtls dr guard to correct block

incfly:


howardjohn (approved):

incfly:
The problem of no diff is because we don't render the `istio-demo-auth.yaml`.

PR #511, opened 2019-11-06
automtls=true global=true case in 1.4

incfly:

### DIFF
--- a/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
+++ b/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
@@ -18,8 +18,8 @@ spec:
   peers:
   - mtls: {}
 ---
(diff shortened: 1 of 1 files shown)


istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
```diff
@@ -33,6 +33,7 @@ spec:
     tls:
       mode: ISTIO_MUTUAL
 ---
+{{ end }}
```
diemtvu: I think this should cover both DRs, i.e move this `{{end}}` to the end of the next DR.

incfly: sure. though this creates some delta between 1.4 and master then.

howardjohn (approved):

PR #524, opened 2019-11-08
Port subscribedResources from old charts

ostromart:


howardjohn (approved):

ostromart:
Are these timeouts flakes or could something else be going on? 
I think these tests passed the first time I ran them (gencheck failed though).

istio-control/istio-discovery/templates/configmap.yaml
```diff
@@ -183,6 +183,19 @@ data:
     certificates:
 {{ toYaml .Values.global.certificates | trim | indent 6 }}
 
+    {{- if $.Values.global.useMCP }}
+    configSources:
```
ostromart: @elfinhe if we look at line 163 there is config_sources already set. Which of these is correct?
This PR causes tests to fail so I suspect the previous setting was not doing anything.

elfinhe: I do not know, but I prefer to believing `configSources` is correct because it is something that has been used for a while. This MCP configSources is defined [here](https://github.com/istio/api/blob/master/mesh/v1alpha1/config.pb.go#L313).

elfinhe: By say failed tests, are we talking about this timeout below:
`error: timed out waiting for the condition on deployments/istio-pilot`?

elfinhe: @ayj please take a look. I guess we need to change the template condition to 
```

    {{- if .Values.global.useMCP }}
    configSources:
    {{- if .Values.global.controlPlaneSecurityEnabled}}
    - address: istio-galley.{{ .Values.global.configNamespace }}.svc:15019
    {{- else }}
    - address: istio-galley.{{ .Values.global.configNamespace }}.svc:9901
    {{- end }}
      {{- if .Values.pilot.configSource.subscribedResources }}
      subscribedResources:
        - {{ .Values.pilot.configSource.subscribedResources }}
      {{- end}}
    {{- if $.Values.global.controlPlaneSecurityEnabled}}
      tlsSettings:
        mode: ISTIO_MUTUAL
    {{- end }}
    {{- end }}

```

ostromart: Yes, there's some configuration problem with pilot with the config in this PR. I thought it was a flake at first but it fails every time.

kustomize/istio-canary/discovery.gen.yaml
```diff
@@ -194,6 +194,8 @@ data:
     autoscaleMin: 1
     configMap: true
     configNamespace: istio-config
+    configSource:
+      subscribedResources: null
```
howardjohn: this seems likely the problem?

elfinhe:
This is not required and was tested before 1.4 release.

PR #534, opened 2019-11-09
Set controlPlaneSecurityEnabled to false by default

pitlv2109:
All operator profiles, except sds, should set controlPlaneSecurityEnabled to false since they don't need to see security features.


howardjohn (dismissed):

howardjohn:
What is the motivation for this change?

pitlv2109:
> What is the motivation for this change?

Added

costinm:
/hold
By default we ship with 'permissive' enabled, meaning users expect TLS to be used in data plane where possible. 

Both 'secure' and 'permissive' mode - with TLS in data plane - MUST also have control plane security enabled, it is insecure otherwise ( MITM of control plane can result in envoy config set to arbitrary values, bypassing any tls in data plane)

howardjohn:
> Both 'secure' and 'permissive' mode - with TLS in data plane - MUST also have control plane security enabled, it is insecure otherwise ( MITM of control plane can result in envoy config set to arbitrary values, bypassing any tls in data plane)

@costinm where did this requirement come from? As far as I know, istio has historically had plain text for the control plane

costinm:
Not sure what is the source of the requirement - common sense maybe :-)

If we are using JWT tokens (bearer) when authenticating to control plane for the new SDS - the OAuth standard does require bearer tokens to be sent over encrypted transport AFAIK. 

Also if we are in strict mode - we can't claim to be secure if the control plane can be MITM, since it
is serving all the security policies. 

For permissive - pretty much same issue, the assumption is that some workloads will be secured
and use auto-mtls - but if the control plane is not secured, the workloads can't be considered secure either.

PR #541, opened 2019-11-11
update for telemetryv2

richardwxn:
mixerreportserver still need to be on now for tcp

### DIFF
--- a/istio-control/istio-discovery/templates/configmap.yaml
+++ b/istio-control/istio-discovery/templates/configmap.yaml
@@ -50,7 +50,7 @@ data:
     {{- else }}
     mixerCheckServer: istio-policy.{{ .Values.global.policyNamespace }}.svc.cluster.local:9091
     {{- end }}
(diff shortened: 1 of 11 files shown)


istio-control/istio-discovery/values.yaml
```diff
@@ -105,7 +105,7 @@ pilot:
 
     v1:
       # Will define mixerCheckServer and mixerReportServer
-      enabled: false
+      enabled: true
```
howardjohn: isn't just this change sufficient?

mandarjog: From an API perspective, we just want to say `v2.enabled: true`  at present part of v1 is implemented by `v1` but this is implementation detail.

This should still be `false`

richardwxn: done, PTAL

howardjohn (approved):

PR #542, opened 2019-11-12
add namespace for envoyfilter.

### LINKED ISSUE #18878: Operator controller gets reconciling error: Namespace parameter required.
**Bug description**

With the latest master, the operator controller always complain the error `Namespace parameter required.` while updateing the resources, result in incorrect ICP status:
```
status:
  status:
    Base:
      status: 2
      statusString: HEALTHY
    Citadel:
      status: 2
      statusString: HEALTHY
    Galley:
      status: 2
      statusString: HEALTHY
    Grafana:
      status: 2
      statusString: HEALTHY
    IngressGateway:
      status: 2
      statusString: HEALTHY
    Injector:
      status: 2
      statusString: HEALTHY
    Pilot:
      error: Namespace parameter required.
      status: 3
      statusString: ERROR
    Policy:
      status: 2
      statusString: HEALTHY
    Telemetry:
      status: 2
      statusString: HEALTHY
    Tracing:
      status: 2
      statusString: HEALTHY
```

**Expected behavior**

**Version (include the output of `istioctl version --remote` and `kubectl version`)**

**How was Istio installed?**

morvencao:
resolve: https://github.com/istio/istio/issues/18878


howardjohn (approved):
this technically should be the root namespace which may not be the same as release. I think global.rootNamespaceor something

istio-control/istio-discovery/metadata-exchange-v2.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: networking.istio.io/v1alpha3
 kind: EnvoyFilter
 metadata:
   name: metadata-exchange
+  namespace: {{ .Release.Namespace }}
```
mandarjog: Is this the config namespace? We want these to apply to all namespaces.

morvencao:
Looks like the imported files will not be send to helm render, so the added `namespace: {{ .Release.Namespace }}` will not be rendered correctly.

istio-control/istio-discovery/metadata-exchange-v2.yaml
morvencao: should it be in `global.configNamespace`?

richardwxn:
but is EnvoyFilter scoped to namespace?

morvencao:
Yes, envoyfiler is namespace scoped resource.

I have added `global.configNamespace` to the two config files.

howardjohn:
configNamespace is not right, that is where galley lives. Our whole namespace setup isn't great, but we need this code from istio/istio in the mesh config -- right now it is missing:

https://github.com/istio/istio/blob/2b971c71aa807f777219c1bf64ed6932770d6f06/install/kubernetes/helm/istio/templates/configmap.yaml#L157-L161

morvencao:
@howardjohn Would you take another look about the change set? I have updated the PR.

howardjohn (approved):

howardjohn:
@morvencao I approved this, not sure if you want the hold still

PR #545, opened 2019-11-12
add grafana storage size

yacut:
related to https://github.com/istio/istio/pull/18739


morvencao (commented):
/ok-to-test

sdake:
@yacut Similar in vein to the comments here https://github.com/istio/installer/pull/546#issuecomment-558066361 I think we should get out of the third party install business and into the third party integration business (via their operators).

Cheers
-steve

PR #549, opened 2019-11-12
Move telemetry v2 values to telemetry.v2

howardjohn:
From pilot.telemetry.v2

For https://github.com/istio/istio/issues/18881

PR #550, opened 2019-11-12
Move telemetry v2 values

howardjohn:
Cherrypick of https://github.com/istio/installer/pull/549/files

### DIFF
--- a/istio-control/istio-discovery/templates/configmap.yaml
+++ b/istio-control/istio-discovery/templates/configmap.yaml
@@ -50,7 +50,7 @@ data:
     {{- else }}
     mixerCheckServer: istio-policy.{{ .Values.global.policyNamespace }}.svc.cluster.local:9091
     {{- end }}
-    {{- if and .Values.telemetry.enabled}}
+    {{- if .Values.telemetry.enabled}}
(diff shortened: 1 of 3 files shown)


istio-control/istio-discovery/templates/configmap.yaml
```diff
@@ -50,7 +50,7 @@ data:
     {{- else }}
     mixerCheckServer: istio-policy.{{ .Values.global.policyNamespace }}.svc.cluster.local:9091
     {{- end }}
-    {{- if and .Values.pilot.telemetry.enabled .Values.pilot.telemetry.v1.enabled }}
+    {{- if and .Values.telemetry.enabled}}
```
jwendell: remove `and`?

sdake: @jwendell I am not sure what the desired approach here was, but master should be fixed first if your suggestion is correct.

jwendell (approved):

istio-control/istio-discovery/templates/configmap.yaml
jwendell: @sdake it's fixed in master. I think that was a typo/leftover when backporting to 1.4.

PR #552, opened 2019-11-14
Fix sds bootstrap configs for Istio control plane

pitlv2109:
* Fix SDS bootstrap config for Pilot, Galley, and Mixer. Prior to this, even if sds is used, aforementioned components still used file mount certs. This will break Istio if the control plane is set to use an SDS-based CA (pods like ingress gateway will fail to connect to Pilot, etc. due to having different root certs).

* Configure the Mixer checker server to use sidecar for MCP connections to align with the rest of Istio control plane components.


pitlv2109:
Please cc relevant people from Pilot, Galley, and Mixer. I tested it locally with security features and everything looks fine.

myidpt (approved):

istio-policy/templates/deployment.yaml
```diff
@@ -95,7 +98,7 @@ spec:
 {{- end }}
 {{- if .Values.global.useMCP }}
     {{- if .Values.global.controlPlaneSecurityEnabled}}
-          - --configStoreURL=mcps://istio-galley.{{ .Values.global.configNamespace }}.svc:15019
+          - --configStoreURL=mcp://localhost:15019
```
JimmyCYJ: Is this change consistent with the old installer? I am not clear why this is needed.

JimmyCYJ (approved):

istio-policy/templates/deployment.yaml
pitlv2109: This is for the policy mixer container to talk to its sidecar, similar to https://github.com/istio/installer/pull/108. In the old installer, Galley doesn't have a sidecar and apparently policy-mixer bypasses its sidecar and talks to Galley directly.

sdake (changes requested):
this is a pretty big change for a bunch of yaml :).  What was the source of these changes?  Manually generated, or came from some other repository?

Cheers
-steve

istio-control/istio-config/templates/deployment.yaml
```diff
@@ -181,9 +181,28 @@ spec:
...
+      {{- if .Values.global.sds.enabled }}
+      - hostPath:
+          path: /var/run/sds
+        name: sds-uds-path
+      - name: istio-token
+        projected:
+          sources:
+          - serviceAccountToken:
+              audience: {{ .Values.global.sds.token.aud }}
+              expirationSeconds: 43200
```
sdake: why this expirationSeconds magic number?

istio-control/istio-config/templates/configmap-envoy.yaml
```diff
@@ -74,6 +74,54 @@ data:
                     route:
                       cluster: in.9901
                       timeout: 0.000s
+        {{- if .Values.global.sds.enabled }}
+          tls_context:
+            common_tls_context:
+              alpn_protocols:
+              - h2
```
sdake: `h2` is some type of alpn protocol?

istio-control/istio-config/templates/deployment.yaml
```diff
@@ -181,9 +181,28 @@ spec:
             readOnly: true
           - name: envoy-config
             mountPath: /var/lib/istio/galley/envoy
+          {{- if .Values.global.sds.enabled }}
+          - name: sds-uds-path
+            mountPath: /var/run/sds
+            readOnly: true
+          - name: istio-token
+            mountPath: /var/run/secrets/tokens
```
sdake: will this not collide with K8s's `/var/run/secrets/tokens` mount?

istio-control/istio-config/templates/deployment.yaml
```diff
@@ -181,9 +181,28 @@ spec:
...
+          {{- end }}
 {{- end }}
 
       volumes:
+      {{- if .Values.global.sds.enabled }}
+      - hostPath:
```
sdake: why a hostpath?  There is no good reason to use a host path in anywhere of the manifests of Istio with the possible exception of CNI.

istio-control/istio-discovery/templates/configmap-envoy.yaml
```diff
@@ -47,7 +47,57 @@ data:
...
+                          config:
+                            header_key: istio_sds_credentials_header-bin
+                            secret_data:
+                              filename: {{ .Values.global.sds.token.path }}
+                      credentials_factory_name: envoy.grpc_credentials.file_based_metadata
+                      stat_prefix: sdsstat
+            combined_validation_context:
+              default_validation_context:
+                verify_subject_alt_name:
+                - spiffe://{{ .Values.global.trustDomain }}/ns/{{ .Values.global.configNamespace }}/sa/istio-galley-service-account
```
sdake: duplicated at 112...

istio-control/istio-discovery/templates/configmap-envoy.yaml
```diff
@@ -47,7 +47,57 @@ data:
...
+                          local_credentials: {}
+                        call_credentials:
+                        - from_plugin:
+                            name: envoy.grpc_credentials.file_based_metadata
+                            config:
+                              header_key: istio_sds_credentials_header-bin
+                              secret_data:
+                                filename: {{ .Values.global.sds.token.path }}
+                        credentials_factory_name: envoy.grpc_credentials.file_based_metadata
+                        stat_prefix: sdsstat
```
sdake: no underscore in sdsstat?

sdake (changes requested):
once the hostpath problem is fixed (or explained in great detail) - and the other nitpicks are addressed or confirmed as accurate, this (complex to review) PR should be ready to go.  I don't have 100% confidence this code is functional - we can always to a partial or full revert if need be.

sdake:
I'd prefer not to cherrypick this right away (or approve the cherrypick right away) for 1.4.  This is a large change that is not suitable for a backport.  I understand the significance of missing SDS, so willing to defer to the release managers on 1.4 for this PR, however, I really think it needs some baking time in master for a few weeks before backporting given its heft.

pitlv2109:
> this is a pretty big change for a bunch of yaml :). What was the source of these changes? Manually generated, or came from some other repository?
> 
> Cheers
> -steve

The bootstrap configurations are from the old installer (e.g https://github.com/istio/istio/tree/master/pilot/docker).

istio-control/istio-discovery/templates/configmap-envoy.yaml
pitlv2109: This is not a duplicate. Each one for a if-else branch

istio-control/istio-config/templates/deployment.yaml
pitlv2109: From the old installer and this is what we've been using. I'd rather not changing any constants in this PR.

global.yaml
```diff
@@ -443,9 +443,11 @@ global:
     enabled: false
     udsPath: ""
     # The JWT token for SDS and the aud field of such JWT. See RFC 7519, section 4.1.3.
-    # When a CSR is sent from Citadel Agent to the CA (e.g. Citadel), this aud is to make sure the
-    # JWT is intended for the CA.
     token:
+      # Filepath of the token.
+      path: ""
```
costinm: Why empty ? The feature is enabled by sds.enabled, no need to have to configure the path too.

istio-control/istio-config/templates/configmap-envoy.yaml
```diff
@@ -74,6 +74,54 @@ data:
                     route:
                       cluster: in.9901
                       timeout: 0.000s
+        {{- if .Values.global.sds.enabled }}
```
costinm: Pilot has support for K8S signed certificates - which are needed for inject, etc.
And the feedback in ToC was to remove the sidecar if I remember correctly.

costinm (commented):
For 1.5 - this conflicts with the plan to use K8S-signed certs and use istiod mode ( injector, etc in pilot ). 

For 1.4 - this seems far too large change. 

For both - it's quite complicated, and we already have a lot of problems with helm template complexity.

Better approach: focus on using SDS in pilot-agent and turn on SDS-only in istiod, it doesn't require any installer change except mounting the token (which can be optional, so no options), it's all in the code.

istio-control/istio-config/templates/deployment.yaml
myidpt: This is 12 hours. We made this the default JWT lifetime. I think it's a balance of making the JWT short-lived VS. not causing too much JWT rotation overhead.

istio-control/istio-config/templates/deployment.yaml
myidpt: Yes we agree. The reason for doing that is in we are still using the node agent DaemonSet for SDS. We are working on the per-pod Istio Agent to replace the node agent, which will get rid of the hostPath.

istio-control/istio-config/templates/deployment.yaml
myidpt: Good question. This one is to load the trustworthy JWT to the "tokens" path. The default K8s JWT is in the "kubernetes.io/serviceaccount" path. I don't see they collide since they are using different sub-directories.

PR #554, opened 2019-11-15
Kill gen

howardjohn:


richardwxn (commented):
Are we going to still keep the kustomize directory and the corresponding kustomization files but just remove the gen files

howardjohn:
@richardwxn yes they are generate at test time

howardjohn:
Maybe not long term

sdake (approved):
LGTM.

richardwxn:
@howardjohn you can also remove the whole run-build-default target

costinm:
Not very happy with this - kubectl apply -k without downloads was very convenient and is what K8S recommends for installs.

I will continue to maintain kustomize templates in a private repo for now. 

Keep in mind people use CI/CD and other ways to install istio - and we want to move away from helm dependency ( operator is just using helm with same complex templates ).

PR #557, opened 2019-11-17
remove egress sds

richardwxn:
we not yet support egress sds, right? need confirmation from @incfly @pitlv2109

### DIFF
--- a/gateways/istio-egress/templates/deployment.yaml
+++ b/gateways/istio-egress/templates/deployment.yaml
@@ -197,12 +197,6 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-          {{- if $gateway.sds }}
-          {{- if $gateway.sds.enabled }}
-          - name: ISTIO_META_USER_SDS
-            value: "true"
-          {{- end }}
-          {{- end }}
           {{- if $gateway.env }}
           {{- range $key, $val := $gateway.env }}
           - name: {{ $key }}
--- a/gateways/istio-egress/values.yaml
+++ b/gateways/istio-egress/values.yaml
@@ -56,20 +56,6 @@ gateways:
       secretName: istio-egressgateway-ca-certs
       mountPath: /etc/istio/egressgateway-ca-certs
 
-    sds:
-      # If true, ingress gateway fetches credentials from SDS server to handle TLS connections.
-      enabled: false
-      # SDS server that watches kubernetes secrets and provisions credentials to ingress gateway.
-      # This server runs in the same pod as ingress gateway.
-      image: node-agent-k8s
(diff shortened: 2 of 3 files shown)


gateways/istio-egress/templates/deployment.yaml
```diff
@@ -197,12 +197,6 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-          {{- if $gateway.sds }}
-          {{- if $gateway.sds.enabled }}
```
incfly: `gateway.sds.enabled` used for public facing certificate for ingress. From the doc I can tell, https://istio.io/docs/tasks/traffic-management/ingress/secure-ingress-sds/

We do support SDS, for egress gateway, for workload identity, between workload and egress, but that's governed already by `if .Values.global.sds.enabled`  below.

I can't think of a use case to support https cert on egress, seems abnormal. 

@JimmyCYJ WDTY

incfly:
Does this cause any problem/failures, If just clean up, then might not worth the risk?

richardwxn:
@incfly It is a clean up because it doesn't seem to be used

pitlv2109:
I don't think we have sds for egress (to provision and exchange certs with external services), but @JimmyCYJ can confirm.

gateways/istio-egress/values.yaml
```diff
@@ -56,20 +56,6 @@ gateways:
       secretName: istio-egressgateway-ca-certs
       mountPath: /etc/istio/egressgateway-ca-certs
 
-    sds:
-      # If true, ingress gateway fetches credentials from SDS server to handle TLS connections.
-      enabled: false
-      # SDS server that watches kubernetes secrets and provisions credentials to ingress gateway.
```
pitlv2109: Maybe this is a copy paste error from the ingress, as you can see all the comments are for ingress

gateways/istio-egress/templates/deployment.yaml
JimmyCYJ: @incfly We don't support SDS for external traffic at egress. There is a tracking GH issue https://github.com/istio/istio/issues/14039.

incfly (approved):

gateways/istio-egress/templates/deployment.yaml
incfly: Thanks confirmation. Approving.

incfly (approved):

richardwxn:
looks like the helm lint is failing for no entry for map for new PRs, I think this was fixed long time ago in helm, might be a helm lint regression?

morvencao (approved):
Approve this since all comments has been addressed.

PR #562, opened 2019-11-20
Fix helm template option in iop

110y:
In addition to https://github.com/istio/installer/pull/558, fixed `helm template` option in `iop`.


richardwxn (commented):
why not just use --name

110y:
@richardwxn 

In `helm` 3.0.0, `--name` option is no longer available.

richardwxn (approved):

PR #573, opened 2019-11-26
Resolution for a regression

sdake:
Fixes: #19227

This PR mirrors: https://github.com/istio/istio/pull/19231

I am not confident this change is correct, however, it does revert back to the original
1.3 behavior. I don't understand why we want two services (zipkin and tracing). Should not
tracing be sufficient?

If we must retain zipkin and we want tracing to run on port 9411 (vs what it has been running
on which is port 80...) zipkin must run on some other port.

### DIFF
--- a/istio-telemetry/tracing/templates/service.yaml
+++ b/istio-telemetry/tracing/templates/service.yaml
@@ -8,7 +8,7 @@ metadata:
     release: {{ .Release.Name }}
 spec:
   ports:
-    - port: {{ .Values.tracing.service.externalPort }}
+    - port: {{ .Values.tracing.zipkin.queryPort }}
       targetPort: {{ .Values.tracing.zipkin.queryPort }}
       protocol: TCP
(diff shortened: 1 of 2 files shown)


douglas-reid:
`tracing` is the service for access the tracing UI. `zipkin` is the service for trace reporting/collection. you could collapse the two, but then you are requiring that a tracing service also provide the `zipkin` endpoint, which isn't strictly necessary.

douglas-reid (approved):

objectiser (approved):
Although there is no problem with the change, it is technically unnecessary due to https://github.com/istio/installer/pull/563. However if you want to retain the `.Values.service.externalPort`, then may be best to revert my change.

sdake:
@objectiser I am not really clear on the advantage of this PR's approach other than it enables NodePort as a service type from a prior PR.  That prior PR had an error, so this fixes that :)

linsun (approved):

PR #584, opened 2019-12-03
add proxy match on version for telemetry v2 filters

richardwxn:


richardwxn:
@mandarjog I am still testing locally but PTAL of overall idea

mandarjog:
I think this looks ok, though the version specific values should not say “*”
Please ensure that you have both positive and negative tests.

istio-control/istio-discovery/values.yaml
```diff
@@ -138,3 +138,7 @@ telemetry:
     # For Null VM case now. If enabled, will set disableMixerHttpReports to true and not define mixerReportServer
     # also enable metadata exchange and stats filter.
     enabled: false
+
+    proxyMatch:
+      # regular expression to be used to select proxies to apply telemetry v2 envoy filters. e.g. "1.5*"
+      proxyVersion: "*"
```
howardjohn: Shouldn't this be `.*`?

And above would need to be `1\.5.*`?

mandarjog:
@richardwxn please ensure that we emit a proper error, afaik we are silently ignoring the error and then it means “match anything”

istio-control/istio-discovery/values.yaml
richardwxn: yep it should be, I forgot to update

richardwxn:
@mandarjog Can you give an example of the case when error is silenced. Actually if you provide a bad type of the proxyMatch field it would just fail validation and does not apply. But if you supply a regular expression A(assume the format is correct) although it is not you intend to do it would just match A but would not match anything

PR #590, opened 2019-12-04
add stackdriver envoyfilter for telemetryv2

richardwxn:


richardwxn:
Do we need to enable v2 by default? @mandarjog @bianpengyuan

istio-control/istio-discovery/values.yaml
```diff
@@ -137,4 +137,5 @@ telemetry:
   v2:
     # For Null VM case now. If enabled, will set disableMixerHttpReports to true and not define mixerReportServer
     # also enable metadata exchange and stats filter.
-    enabled: false
+    enabled: true
+    sdfilter: true
```
bianpengyuan: Definitely not enable sd filter by default. What does `telemetry.enabled` and `telemetry.v2.enabled` control respectively? I'm trying to figure out what is the right layer for this option... Also let's use full name for Stackdriver (stackdriverFilter).

richardwxn: the outmost layer is gated by telemetry.enabled and telemetry.v2.enabled together now, I think it should be ok to put stackdriverfilter option under v2 as it is

howardjohn: `filter` is an implementation detail. I would suggest just stackdriver

howardjohn: What if I just want stackdriver, no prometheus?

bianpengyuan: Yeah, just talked to Xinnan offline, he is going to add a new option here (stats) to selectively enable stats filter.

howardjohn: The "stats filter" is sort of an implementation detail though. A user just knows they want telemetry. Setting `telemetry=enabled,telemetry.stats=disabled` feels weird, since stats is sort of a synonym for telemetry? 

Is the stats filter just prometheus? If so would it make sense to put this behind an option `prometheus`?

bianpengyuan: make sense to me. @mandarjog your call :)

richardwxn: you are right, stats is implementation detail, prometheus makes more sense actually.

istio-control/istio-discovery/templates/telemetryv2.yaml
```diff
@@ -122,5 +122,89 @@ spec:
...
+  {{- if .Values.global.configRootNamespace }}
+  namespace: {{ .Values.global.configRootNamespace }}
+  {{- else }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
+spec:
+  configPatches:
+    - applyTo: HTTP_FILTER
+      match:
+        context: SIDECAR_OUTBOUND
```
mandarjog: This needs to use proxymatch for 1.4

istio-control/istio-discovery/templates/telemetryv2.yaml
```diff
@@ -122,5 +122,89 @@ spec:
...
+          config:
+            config:
+              root_id: stackdriver_outbound
+              configuration: |
+                {}
+              vm_config:
+                vm_id: stackdriver_outbound
+                runtime: envoy.wasm.runtime.null
+                code:
+                  local: { inline_string: envoy.wasm.null.stackdriver }
```
mandarjog: This is not right for 1.4

https://github.com/istio/proxy/blob/release-1.4/extensions/stackdriver/testdata/stackdriver_filter.yaml

Please look at this for reference. @bianpengyuan

istio-control/istio-discovery/templates/telemetryv2.yaml
```diff
@@ -122,5 +122,89 @@ spec:
...
+                name: "envoy.router"
+      patch:
+        operation: INSERT_BEFORE
+        value:
+          name: envoy.filters.http.wasm
+          config:
+            config:
+              root_id: stackdriver_inbound
+              configuration: |
+                {}
```
mandarjog: configuration should have `enable_mesh_edges_reporting: true` set.

istio-control/istio-discovery/templates/telemetryv2.yaml
richardwxn: I would do the proxyMatch part in this PR: https://github.com/istio/installer/pull/584

istio-control/istio-discovery/templates/telemetryv2.yaml
richardwxn: do we need the local part? the master and release-1.4 has discrepancy here

istio-control/istio-discovery/values.yaml
mandarjog: yes, lets call it

```yaml
stackdriver:
   enabled: true
   logging: true
   monitoring: true
   topology: true
 prometheus:
   enabled: true
```

mandarjog: @douglas-reid @bianpengyuan ?

douglas-reid: that lgtm.

istio-control/istio-discovery/templates/telemetryv2.yaml
richardwxn: I would create two PRs, one for master another one for 1.4 because of the filter differences.

istio-control/istio-discovery/values.yaml
richardwxn: is stackdriver all enabled?

douglas-reid: wait. I think the naming here is fine, but I don't think we can enable topology by default.

richardwxn:
@mandarjog @bianpengyuan PTAL this master one
1. 1.4 and 1.5 envoyfilters added with corresponding filter definition and proxy match version
2. default v2.enabled and v2.prometheus.enabled as true, v2.stackdriver.enabled=false (Please confirm this part)

mandarjog (approved):
/lgtm

howardjohn (approved):

PR #591, opened 2019-12-05
[release-1.4] update telemetryv2 templates

richardwxn:
This change is for 1.4 specifically, including the filter part, proxymatch verison part


istio-control/istio-discovery/templates/telemetryv2.yaml
```diff
@@ -1,5 +1,234 @@
 {{- if and .Values.telemetry.enabled .Values.telemetry.v2.enabled }}
-{{ .Files.Get "metadata-exchange-v2.yaml" }}
+apiVersion: networking.istio.io/v1alpha3
```
mandarjog: This should be enabled if either stackdriver or prom is enabled.

richardwxn: then there is no meaning of the v2.enabled setting? also if we have some more other filters in the future then this check would become lengthy

mandarjog: Ok, that makes sense, this is only enabled when v2 is enabled .

istio-control/istio-discovery/templates/telemetryv2.yaml
```diff
@@ -1,5 +1,234 @@
...
+  {{- else }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
+spec:
+  configPatches:
+    - applyTo: HTTP_FILTER
+      match:
+        context: ANY # inbound, outbound, and gateway
+        proxy:
+          proxyVersion: {{ .Values.telemetry.v2.proxyMatch.proxyVersion }}
```
bianpengyuan: @mandarjog  How are we going to deal with upgrade? Another set of envoyfilter for post 1.4 proxy version?

richardwxn: the envoyfilter templates are compiled in together with istioctl so they are versioned under the hood already? @ostromart WDYT?

howardjohn: This is not how it will work in practice. The envoyfilter is global, but during an upgrade you will have multiple versions of a proxy running which is where it breaks

richardwxn: but if the proxyMatch part we can still select proxies whose versions are compatible with the envoyFilters?

bianpengyuan: Yeah, you will need to make both envoyfilter 1.4 and 1.5 installed to not break user.

mandarjog: We have 2 choices.
1. Encode the version in the name, so those Configs can coexist.
2. Leave it as is, but the next version of filters configs will have extra config patches of the previous version. After the api stabilizes I don’t expect we will need this so much.

If we go with option 1, this pr needs to be updated.

richardwxn: I am ok with first option

mandarjog (approved):
/lgtm

istio-control/istio-discovery/values.yaml
```diff
@@ -124,4 +124,21 @@ telemetry:
   v2:
     # For Null VM case now. If enabled, will set disableMixerHttpReports to true and not define mixerReportServer
     # also enable metadata exchange and stats filter.
-    enabled: false
\ No newline at end of file
+    enabled: true
```
howardjohn: Can you add the master/1.5 filters? Otherwise we can't use telemetry v2 on master yet we are enabling it..

mandarjog: That should go in the master branch, not 1.4

howardjohn: Oh I thought this was master.

In that case this should be false

richardwxn: put it back as force

howardjohn:
We need to have 1.4 have this, and master have 1.5+1.4. Otherwise when we upgrade its just going to prune this one, right?

richardwxn:
@howardjohn yes I updated the master PR as well: https://github.com/istio/installer/pull/590. Plz check

bianpengyuan (approved):

istio-control/istio-discovery/values.yaml
```diff
@@ -124,4 +124,21 @@ telemetry:
...
     # also enable metadata exchange and stats filter.
-    enabled: false
\ No newline at end of file
+    enabled: false
+    # Indicate if prometheus stats filter is enabled or not
+    prometheus:
+      enabled: false
+    # stackdriver filter settings.
+    stackdriver:
+      enabled: true
```
bianpengyuan: Stackdriver should be disabled by default as well.

richardwxn: I updated the stackdriver configs to be disabled, now then if we want to enable stackdriver filter then we need to do sth like this: 

istioctl manifest apply --set values.telemetry.enabled=true --set values.telemetry.v2.enabled=true --set values.telemetry.v2.stackdriver.enabled=true

A little bit lengthy but less confusing.

richardwxn:
@howardjohn plz check again?

howardjohn:
@richardwxn can we merge master first so we ensure they are in sync? And have telemetry experts approve that one and I will merge

howardjohn (approved):

PR #592, opened 2019-12-05
Configure destination rules in config root namespace

haines:
I have configured `configRootNamespace: istio-config`, but the destination rules are put in `istio-control` where I deployed `istio-discovery`.

Also, none of them seem to be applying, and I'm pretty sure that's because the mesh's root namespace isn't configured and therefore it's falling back to `istio-system` (which doesn't exist).

https://github.com/istio/istio/blob/e921e4b7410a9e0a7c736e96a510266041d9fc7a/install/kubernetes/helm/istio/templates/configmap.yaml#L157-L163

### DIFF
--- a/base/templates/namespaces.yaml
+++ b/base/templates/namespaces.yaml
@@ -30,6 +30,17 @@ metadata:
 ---
 {{- end }}
 
+{{- if and (ne .Values.global.configRootNamespace .Release.Namespace) (ne .Values.global.configRootNamespace .Values.global.configNamespace) }}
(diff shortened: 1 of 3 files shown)


morvencao (commented):
/ok-to-test

haines:
I've rebased this onto the latest master, please let me know if there's anything else I need to do to get it merged!

haines:
Closing, for the same reasons as https://github.com/istio/istio/pull/22779

PR #601, opened 2019-12-10
add tolerations

### LINKED ISSUE #19480: There is no way to provide "tolerations" to Kiali and tracing components
**Bug description**
There is no way to provide "tolerations" to Kiali component

Istio 1.4.1

I've tried the following:

- Specifying values.kiali.tolerations - doesn't work, there is no such field in spec. Check this file:
https://github.com/istio/operator/blob/9f96fae012485e3fbeeac88ca6b1ab4534f67dd7/pkg/apis/istio/v1alpha1/values_types.proto#L1791
BTW, this also means that documentation is wrong, as the field is there:
https://istio.io/docs/reference/config/installation-options/#kiali-options
- Specifying unvalidatedValues.kiali.tolerations via file - doesn't work, as all unvalidatedValues are ignored. Relevant PR for upgrade: https://github.com/istio/operator/pull/671
- Specifying unvalidatedValues.kiali.tolerations - via cmd - doesn't work, as complex values are not supported. Relevant issue: https://github.com/istio/istio/issues/19392
- Specifying through IstioControlPlane is impossible as there is no telemetry.kiali component

Looks like currently there are only three options:

- Deploying via Helm
- Deploying Kiali as a separate chart (not sure if it is possible)
- Adding additional tweak step to alter Kiali pod fields

richardwxn:
To resolve https://github.com/istio/istio/issues/19480

### DIFF
--- a/istio-telemetry/kiali/templates/deployment.yaml
+++ b/istio-telemetry/kiali/templates/deployment.yaml
@@ -90,3 +90,6 @@ spec:
       affinity:
       {{- include "nodeaffinity" . | indent 6 }}
       {{- include "podAntiAffinity" . | indent 6 }}
+      {{- if .Values.kiali.tolerations }}
+      tolerations:
+      {{ toYaml .Values.kiali.tolerations | indent 6 }}
--- a/istio-telemetry/kiali/templates/deployment.yaml
+++ b/istio-telemetry/kiali/templates/deployment.yaml
@@ -93,3 +93,4 @@ spec:
       {{- if .Values.kiali.tolerations }}
       tolerations:
       {{ toYaml .Values.kiali.tolerations | indent 6 }}
+      {{- end }}
--- a/istio-telemetry/kiali/values.yaml
+++ b/istio-telemetry/kiali/values.yaml
@@ -10,6 +10,7 @@ kiali:
   contextPath: /kiali # The root context path to access the Kiali UI.
   nodeSelector: {}
   podAnnotations: {}
+  tolerations: []
 
   # Specify the pod anti-affinity that allows you to constrain which nodes
   # your pod is eligible to be scheduled based on labels on pods that are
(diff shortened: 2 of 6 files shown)


howardjohn (commented):
Why not just use the k8s spec?

Seems like we are missing an `{{ end }}`?

richardwxn:
@howardjohn yep we should actually use the k8s specs under IstioControlPlane, but there are users complain that we claim to still support helm api at this stage so they tend to believe that overlaying k8s settings under values passthrough part would work as the same as in IstioControlPlane spec.

As an alternative, I have a change locally to improve the validation function, for example it should give a warning and suggests the correct path when user try to set in values pass through

nick4fake:
@howardjohn 

> Why not just use the k8s spec?
Because there is no way to specify it through k8s spec.

howardjohn:
@nick4fake seems like its in the proto: https://github.com/istio/operator/blob/8569a17c89f10190d5e799c3e10b9426c18883b3/pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto#L512 does it not work?

richardwxn:
@howardjohn I think the problem is that it is not defined in the templates of kiali and tracing, unless they do a k8soverlay

jmazzitelli:
> I think the problem is that it is not defined in the templates of kiali

The Kiali integration will support these toleration settings once these two installer/operator PRs are able to get merged ;-)

* istio/installer [PR 556](https://github.com/istio/installer/pull/556):
https://github.com/istio/installer/pull/556/files#diff-44b3244d775b916bfc710c4dd80b433bR34
https://github.com/istio/installer/pull/556/files#diff-4fc69d9eb27e507429323cb933f7ad3bR16
https://github.com/istio/installer/pull/556/files#diff-d2ae3febaa0204735dffb73c835471ddR62

* istio/operator [PR 662](https://github.com/istio/operator/pull/622):
https://github.com/istio/operator/pull/622/files#diff-833bbe66939c20c7f9ce787aafc95e06R708

richardwxn:
we can unblock it for tracing and kiali toleration first

richardwxn:
PTAL @howardjohn

PR #605, opened 2019-12-12
Allow set lifecycle to sidecar template

k0da:
Signed-off-by: Dinar Valeev <k0da@opensuse.org>


richardwxn (commented):
You may need to add lifecycle to the values.yaml as well

k0da:
> You may need to add lifecycle to the values.yaml as well
done

k0da:
@richardwxn any other review comments?

richardwxn:
for your values.yaml file definition, I think the lifecycle section is a map not an array?

k0da:
@richardwxn right, lifecycle is a map

richardwxn (approved):

PR #606, opened 2019-12-13
Initial merging of istiod into mainstream templates

costinm:
This should run istiod in SDS mode and have all functionality from 'standalone' istiod.

However the injection will still use the old injector, and tests still point to the old port - different PR to take care of this. The goal is to switch most of the testing to the 'SDS by default' path which is the
riskier part of istiod.


istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -0,0 +1,373 @@
+# Configmap optimized for Istiod. Please DO NOT MERGE all changes from istio - in particular those dependent on
+# Values.yaml, which should not be used by istiod.
+
+# Istiod only uses SDS based config ( files will mapped/handled by SDS).
```
howardjohn: this file is reverting a few fixes that landed. Looks like the intent was to remove just SDS and add CA_ADDR but it did more?

global.yaml
```diff
@@ -63,6 +63,10 @@ global:
     # will result in LDS rejection and the ingress will not work.
     enableHttps: false
 
+  # Enabled by default in master for maximising testing.
+  istiod:
```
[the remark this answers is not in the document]
nmittler: Are there any major use cases that istiod does not support today?

istio-control/istio-discovery/files/injection-template.yaml
nmittler: This is a little concerning. We should try to find a way to share between istio and istiod.  Maybe istio could read this file and patch it?

istio-control/istio-discovery/templates/clusterrole.yaml
```diff
@@ -48,3 +48,100 @@ rules:
   verbs: ["get", "list", "watch"]
 ---
 {{ end }}
+
+{{ if .Values.global.istiod.enabled }}
+# Dedicated cluster role - istiod will use fewer dangerous permissions ( secret access in particular ).
```
nmittler: nit: this comment could probably use some love, since the context will quickly get lost going forward. Might want to expand on what you mean by "dangerous permissions" and also answer the question "fewer than what?".

global.yaml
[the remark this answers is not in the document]
costinm: @howardjohn  I don't care either, but seems consistent with the others.  Given the mess in values.yaml - don't want to argue the name, anything is fine. 

@nmittler  - we are finding some 'old k8s' use cases, there are solutions but may not be ready in 1.5. However for 'modern' k8s we should be good.

istio-control/istio-discovery/files/injection-template.yaml
costinm: The intent is to simplify - we want the istiod config to be as easy as possible, and readable by admins. And certainly not using any values.yaml. 

Sometimes 'sharing' implies complexity, and not sharing is easier. 

Note that we also want to support users bringing their own (clean) injection templates. We know users who have this use case ( like Google, or others not using envoy, etc ). This will imply some stability in the injector API surface and making the template as minimal as possible - the work on Pilot returning mesh.config and the bootstrap is part of it.

costinm:
PTAL - I'll have a follow-up to merge the inject template (and simplify it further), as well as enabling the injection and other things. Trying to keep it in small pieces.

howardjohn (approved):

PR #612, opened 2019-12-18
Istiod injection

costinm:
Change the inject address to istiod in the webhook. 

Note that the name must match the generated certificate - we can't run Istiod as part of the istio-sidecar-injector service or replace it, since it has a istiod cert.

With this change, injection and SDS will use Istiod and Citadel, injector should no longer get any requests.

Follow up to https://github.com/istio/installer/pull/606


istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -136,12 +136,12 @@ template: |
     - "{{ .ProxyConfig.StatsdUdpAddress }}"
```
costinm: This is merging changes from auto-inject.

howardjohn (approved):

howardjohn (approved):

ayj (approved):

howardjohn:
I have a copy of this with conflict resolved and tests passing if you want to approve that or we can go with this https://github.com/istio/installer/pull/633

istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -136,20 +136,23 @@ template: |
...
     - --proxyAdminPort
     - "{{ .ProxyConfig.ProxyAdminPort }}"
     {{ if gt .ProxyConfig.Concurrency 0 -}}
     - --concurrency
     - "{{ .ProxyConfig.Concurrency }}"
     {{ end -}}
-    {{- if .Values.global.controlPlaneSecurityEnabled }}
+    {{- if .Values.global.istiod.enabled }}
+    - --controlPlaneAuthPolicy
+    - NONE
```
myidpt: Why not TLS?

howardjohn:
I copied this over to #633 which was merged

istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -172,10 +175,14 @@ template: |
...
   {{- end }}
+  {{- if .Values.global.proxy.lifecycle }}
+    lifecycle:
+      {{ toYaml .Values.global.proxy.lifecycle | indent 4 }}
+    {{- end }}
     env:
     # Temp, pending PR to make it default or based on the istiodAddr env
     - name: CA_ADDR
-      value: istiod.istio-system.svc:15012
+      value: istio-pilot.istio-system.svc:15012
```
myidpt: istio-system is hardcoded?

istio-control/istio-discovery/files/injection-template.yaml
howardjohn: I think we are setting it in code https://github.com/istio/istio/blob/8a7495ee55e13db5bf947ebb2c50061cc3c2bf89/pilot/cmd/pilot-agent/main.go#L377, not sure why it matters what it is set to here then though

istio-control/istio-discovery/templates/deployment.yaml
```diff
@@ -141,7 +143,7 @@ spec:
             value: "{{ .Values.pilot.enableProtocolSniffingForInbound }}"
 {{- if .Values.global.istiod.enabled }}
           - name: WEBHOOK
-            value: istiod
+            value: istio-sidecar-injector
```
myidpt: Just curious: why do you make this change?

howardjohn: There is no webhook named "istiod". The webhook name is istio-sidecar-injector. see https://github.com/istio/istio/issues/19780

PR #620, opened 2019-12-23
Fix 1.5 telemetry v2 config

howardjohn:


istio-control/istio-discovery/templates/telemetryv2_1.5.yaml
```diff
@@ -124,13 +121,13 @@ spec:
                 {
                   "debug": "false",
                   "stat_prefix": "istio",
+                  "disable_host_header_fallback": true
```
kyessenov: I think we may want this.

howardjohn: Does want this mean `true` or `false`?

richardwxn:
Are you using the templates from istio/proxy? But the one from istio tests should be the source of truth? https://github.com/istio/istio/blob/master/tests/integration/telemetry/stats/prometheus/testdata/stats_filter.yaml

Need confirmation from @bianpengyuan @douglas-reid @mandarjog

howardjohn:
I copied from istio/proxy. it would be nice to not have this in 4 different places

the main change here is to the metadata exchange the other part is just formatting to match the other examples

PR #625, opened 2019-12-24
Add role for egress SDS

howardjohn:


clyang82:
What is use case for egress gateway fetches credentials from SDS server?

howardjohn:
I think we don't need this - I need to investigate more

/hold

PR #626, opened 2019-12-26
limit cni in kube-system

### LINKED ISSUE #19791: Install istio-cni in namespace istio-cni failed
**Bug description**

Follow this document - https://github.com/istio/installer#install-istio-cni cannot install istio cni. it has such error:
```
Error creating: pods "istio-cni-node-" is forbidden: pods with system-cluster-critical priorityClass is not permitted in istio-cni namespace
```

**Expected behavior**

**Version (include the output of `istioctl version --remote` and `kubectl version` and `helm version` if you used Helm)**

**How was Istio installed?**

clyang82:
Fixed https://github.com/istio/istio/issues/19791

it was resolved in 1.17 - Critical pods should not be limited to kube-system namespace
 https://github.com/kubernetes/kubernetes/issues/60596

### DIFF
--- a/istio-cni/templates/configmap-cni.yaml
+++ b/istio-cni/templates/configmap-cni.yaml
@@ -2,7 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: istio-cni-config
-  namespace: kube-system
+  namespace: {{ .Release.Namespace }}
   labels:
     app: istio-cni
     release: {{ .Release.Name }}
--- a/istio-cni/templates/configmap-cni.yaml
+++ b/istio-cni/templates/configmap-cni.yaml
@@ -2,7 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: istio-cni-config
-  namespace: {{ .Release.Namespace }}
+  namespace: kube-system
   labels:
     app: istio-cni
     release: {{ .Release.Name }}
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ ISTIO_CNI_ARGS=
 if [[ "${ISTIO_CLUSTER_ISGKE}" == "true" ]]; then
     ISTIO_CNI_ARGS="--set cni.cniBinDir=/home/kubernetes/bin"
 fi
-iop istio-cni istio-cni $IBASE/istio-cni/ ${ISTIO_CNI_ARGS}
+iop kube-system istio-cni $IBASE/istio-cni/ ${ISTIO_CNI_ARGS}
 ```
 
 TODO. It is possible to add Istio-CNI later, and gradually migrate.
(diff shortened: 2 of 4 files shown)


clyang82:
we have update the document in this PR - https://github.com/istio/istio.io/pull/4518

howardjohn (commented):
I don't think this is right, shouldn't we just make release.namespace==kube-system? otherwise it can't be customized?

howardjohn:
I think it will actually make the documentation you linked no longer
correct as well

rlenglet:
LGTM. istio-cni must be deployed in kube-system due to the daemonset's priority class. K8s doesn't give any flexibility for high priority classes. The Helm install instructions did the right thing. Maybe that setup got lost in the new installer.

howardjohn:
Why not keep it customizable but set the default to kube system? I think
that's possible. Then the default is correct, but it's consistent with
other components and allows people on latest kubernetes to not install in
kube system

rlenglet:
This is not customizable. K8s requires pods with those priority classes to
run in kube-system. It won't start if this is started in any other
namespaces.

All other CNI plugins deployed as daemonsets run in kube-system for the
same reason. This is consistent.

howardjohn:
according to the link that's no longer true in 1.17

rlenglet:
Oh sorry got it. I'm good with keeping it configurable but default to
kube-system.

clyang82:
the reason is to hardcode `kube-system` without supporting customizable is Istio 1.4 has been tested with Kubernetes releases 1.13, 1.14, 1.15. It is error-prone if we allows configurable. We can enable configurable when we support Kubernetes 1.17.

howardjohn:
This in master branch which is 1.5 and we intend to support 1.17 as far as
I know

howardjohn:
Note that all of our CI runs 1.17 as well

clyang82:
Good to know that. Let me keep it configurable but default to
kube-system

istio-cni/templates/configmap-cni.yaml
```diff
@@ -2,7 +2,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: istio-cni-config
-  namespace: kube-system
+  namespace: {{ .Release.Namespace }}
```
rlenglet: I think it needs its own configuration variable.
When deploying on k8s ≤1.17, this namespace needs to be different from the `Release.Namespace` variable.

howardjohn:
namespace is set per component in this model, it's not the same monolothic
helm chart we had before

clyang82:
Yes. namespace is per-component - `iop kube-system istio-cni $IBASE/istio-cni/ ${ISTIO_CNI_ARGS}`

cagataygurturk:
Hey, at the moment it is impossible to deploy CNI with istioctl because the CNI will be installed to istio-system. I have hopes that this PR would fix it.

richardwxn (commented):
shouldn't we need to update the templates ns as well?

howardjohn:
I think we want to modify data/profiles/default.yaml in operator instead

clyang82:
I am working on that right now.

morvencao (approved):
The minimal doc change is OK according to the discussion.

PR #632, opened 2019-12-31
Support customized labels for istio-ingress

clyang82:
sub-task of https://github.com/istio/istio/issues/19860


morvencao (commented):
You may also need to update the service selector so that it can select the correct pods, see: https://github.com/istio/installer/blob/master/gateways/istio-ingress/templates/service.yaml#L28-L29

morvencao (commented):
do we need to make the same changes to istio-egressgateway?

clyang82:
> You may also need to update the service selector so that it can select the correct pods, see: https://github.com/istio/installer/blob/master/gateways/istio-ingress/templates/service.yaml#L28-L29

Yes. Good catch. Let me update the PR. Thanks.

elfinhe:
error: timed out waiting for the condition on deployments/istio-ingressgateway

elfinhe (approved):

elfinhe:
@morvencao PTAL to double check.

morvencao:
/LGTM
@clyang82 Can you check why the test failed?

morvencao (commented):
/LGTM
@clyang82 Can you check why the test failed?

clyang82:
the failed test was fixed in https://github.com/istio/istio/pull/20041

PR #641, opened 2020-01-07
Disable mTLS for connections to kube-dns.kube-system

### LINKED ISSUE #19968: DNS cannot be resolved via TCP when strict mTLS is enabled
**Bug description**
This seems very similar to #11658, which was marked "resolved", but I don't believe is.  Using a basic install of istio with mTLS enabled, containers cannot resolve DNS via TCP (only via UDP).  This is troublesome because DNS's behavior is to fall back to TCP if the DNS response is large (larger than 512 bytes, I think?).  For example, in a container with strict mTLS enabled, 
```
/ # nslookup -type=srv istio-example-short.reecemath.com
Server:		10.96.0.10
Address:	10.96.0.10#53

Non-authoritative answer:
istio-example-short.reecemath.com	service = 2 10 5000 istio-example2.reecemath.com.
istio-example-short.reecemath.com	service = 1 10 5000 istio-example1.reecemath.com.
```
resolves fine because it uses UDP, but
```
/ # nslookup -type=srv istio-example-long.reecemath.com
;; Truncated, retrying in TCP mode.
;; communications error to 10.96.0.10#53: connection reset
```
does not because it falls back to TCP.

**Expected behavior**
I would expect to be able to use TCP to resolve DNS, but cannot.

1) Download the latest istio (1.4.3 at the time of writing) and install istio with mTLS enabled
```
$ ./istioctl manifest apply --set values.global.mtls.enabled=true --set values.global.controlPlaneSecurityEnabled=true
```
2) Throw a container into istio that has mTLS injection enabled.
```
$ kubectl create ns sleep
$ kubectl label namespace sleep istio-injection=enabled
$ kubectl apply -f ../samples/sleep/sleep.yaml -n sleep
```
3) Shell into the running container
```
$ kubectl get pods -n sleep
NAME                     READY   STATUS    RESTARTS   AGE
sleep-666475687f-675t2   2/2     Running   0          67m
$ kubectl exec -it sleep-666475687f-675t2 -n sleep -- /bin/sh
Defaulting container name to sleep.
Use 'kubectl describe pod/sleep-666475687f-675t2 -n sleep' to see all of the containers in this pod.
/ # dig +short +tcp google.com
;; communications error to 10.96.0.10#53: connection reset
/ # dig +short google.com
172.217.4.78
```

Throwing part of the YAML recommended on https://github.com/istio/istio/issues/11658#issuecomment-475709922 at the cluster "fixes" the problem -- TCP can be used for DNS resolution.  You just need a `DestionationRule` to declare that you shouldn't use TLS for kube-dns.  I currently have this yaml in my cluster build pipeline now as a "workaround", but it seems like a fix within istio's installers would be more ideal.
```
apiVersion: networking.istio.io/v1alpha3
kind: DestinationRule
metadata:
  name: disable-mtls-to-dns
  namespace: kube-system
spec:
  host: "kube-dns.kube-system.svc.cluster.local"
  trafficPolicy:
    tls:
      mode: DISABLE
```

**Version**
```
$ ./istioctl version --remote
client version: 1.4.3
control plane version: 1.4.3
data plane version: 1.4.3 (2 proxies)

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:36:53Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"211047e9a1922595eaa3a1127ed365e9299a6c23", GitTreeState:"clean", BuildDate:"2019-10-15T12:02:12Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```

**How was Istio installed?**
```
./istioctl manifest apply --set values.global.mtls.enabled=true --set values.global.controlPlaneSecurityEnabled=true --set values.global.mtls.auto=false
```

docker-desktop and GCP/GKE

taylorreece:
_I imagine there's a better place to patch this, like in envoy or something._  

See istio/istio#19968 for a write-up of the issue.  TL/DR: mTLS needs to be disabled when accessing kube-dns.kube-system so DNS resolution over TCP can work.

Fixes istio/istio#19968

### DIFF
--- a/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
+++ b/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
@@ -47,6 +47,17 @@ spec:
     tls:
       mode: DISABLE
 ---
+apiVersion: networking.istio.io/v1alpha3
+kind: DestinationRule
+metadata:
+  name: disable-mtls-to-kube-dns
+  namespace: kube-system
+spec:
+  host: "kube-dns.kube-system.svc.cluster.local"
+  trafficPolicy:
+    tls:
+      mode: DISABLE
+---
 {{- else }}
 # Authentication policy to enable permissive mode for all services (that have sidecar) in the mesh.
 apiVersion: "authentication.istio.io/v1alpha1"
--- a/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
+++ b/istio-control/istio-discovery/templates/enable-mesh-mtls.yaml
@@ -33,6 +33,17 @@ spec:
     tls:
       mode: ISTIO_MUTUAL
 ---
+apiVersion: networking.istio.io/v1alpha3
+kind: DestinationRule
+metadata:
+  name: disable-mtls-to-kube-dns
+  namespace: kube-system
+spec:
+  host: "kube-dns.kube-system.svc.cluster.local"
+  trafficPolicy:
+    tls:
+      mode: DISABLE
+---
 {{ end }}
 apiVersion: networking.istio.io/v1alpha3
 kind: DestinationRule
@@ -43,17 +54,6 @@ metadata:
     release: {{ .Release.Name }}
 spec:
   host: "kubernetes.default.svc.{{ .Values.global.proxy.clusterDomain }}"
-  trafficPolicy:
-    tls:
-      mode: DISABLE
----
-apiVersion: networking.istio.io/v1alpha3
-kind: DestinationRule
-metadata:
-  name: disable-mtls-to-kube-dns
-  namespace: kube-system
-spec:
-  host: "kube-dns.kube-system.svc.cluster.local"
   trafficPolicy:
     tls:
       mode: DISABLE


morvencao (commented):
/ok-to-test

howardjohn (commented):
@incfly seems like we shouldn't need this with auto mtls? or does that not apply when strict mode is enabled

taylorreece:
> @incfly seems like we shouldn't need this with auto mtls? or does that not apply when strict mode is enabled

@howardjohn good call.  I hadn't realized https://github.com/istio/installer/commit/3a1b2b9963be7fafe740592597e98f3156a500bd was added so recently.  I moved my change into the "if not .Values.global.mtls.auto" block, as it should only affect things if global.mtls.auto=false.

incfly:
not needed when auto mtls is enabled(regardless of permissive or strict), but could be case auto mtls is disabled. the change now seems right to me.

taylorreece:
@howardjohn with the moving of directories to istio/istio, should I create a new pull request for this against the other istio/istio repo?

taylorreece:
Closing; created a new PR against istio/istio.

PR #645, opened 2020-01-09
Mount Citadel CA cert to data plane and config istiod CA provider

lei-tang:
This PR is for the issue: https://github.com/istio/istio/issues/19950

### DIFF
--- a/istio-control/istio-discovery/files/injection-template.yaml
+++ b/istio-control/istio-discovery/files/injection-template.yaml
@@ -193,7 +193,7 @@ template: |
     {{- end }}
     env:
     - name: SIGN_CERT_AT_KUBERNETES_CA
-      value: {{ .Values.global.signCertAtKubernetesCa }}
+      value: "{{ .Values.global.signCertAtKubernetesCa }}"
     # Temp, pending PR to make it default or based on the istiodAddr env
     - name: CA_ADDR
     {{- if .Values.global.configNamespace }}
@@ -353,7 +353,7 @@ template: |
     {{  end -}}
     volumeMounts:
     {{- if not .Values.global.signCertAtKubernetesCa }}
-    - mountPath: /etc/citadel-ca-cert
+    - mountPath: /etc/istio/citadel-ca-cert
       name: citadel-ca-cert
     {{- end }}
     {{ if (isset .ObjectMeta.Annotations `sidecar.istio.io/bootstrapOverride`) }}
--- a/istio-control/istio-discovery/files/injection-template.yaml
+++ b/istio-control/istio-discovery/files/injection-template.yaml
@@ -192,8 +192,6 @@ template: |
       {{ toYaml .Values.global.proxy.lifecycle | indent 4 }}
     {{- end }}
     env:
-    - name: SIGN_CERT_AT_KUBERNETES_CA
-      value: "{{ .Values.global.signCertAtKubernetesCa }}"
     # Temp, pending PR to make it default or based on the istiodAddr env
     - name: CA_ADDR
     {{- if .Values.global.configNamespace }}
--- a/istio-control/istio-discovery/files/injection-template.yaml
+++ b/istio-control/istio-discovery/files/injection-template.yaml
@@ -192,6 +192,8 @@ template: |
       {{ toYaml .Values.global.proxy.lifecycle | indent 4 }}
     {{- end }}
     env:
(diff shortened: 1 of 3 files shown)


istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -350,6 +352,10 @@ template: |
   {{- end }}
     {{  end -}}
     volumeMounts:
+    {{- if not .Values.global.signCertAtKubernetesCa }}
+    - mountPath: /etc/citadel-ca-cert
```
howardjohn: nit: maybe make this /etc/istio/citadel-ca-cert? so we don't have so much files in random places

howardjohn (commented):
Any reason to not just make it an optional mount and add this unconditionally? Then we don't need to expose a new option

istio-control/istio-discovery/files/injection-template.yaml
lei-tang: Done.

lei-tang:
@howardjohn The reason not add this unconditionally to each proxy is because: 
- .Values.global.signCertAtKubernetesCa is needed anyway as Pilot still needs .Values.global.signCertAtKubernetesCa to configure the SIGN_CERT_AT_KUBERNETES_CA environmental variable.
- The proxy config and Pilot config for SIGN_CERT_AT_KUBERNETES_CA should be consistent.

howardjohn:
> The proxy config and Pilot config for SIGN_CERT_AT_KUBERNETES_CA should be consistent.

You cannot guarantee these will be in sync. Why can't the proxy just use the cert if it exists, and otherwise use DNS cert?

> .Values.global.signCertAtKubernetesCa is needed anyway as Pilot still needs .Values.global.signCertAtKubernetesCa to configure the SIGN_CERT_AT_KUBERNETES_CA environmental variable.

No, its not. Pilot has a default value, which should be a sane default. 99% of users do not need to worry about it. Setting env variables in the deployment is an antipractice in my opinion, we should only do it in cases where pilot can not have a default

istio-control/istio-discovery/files/injection-template.yaml
```diff
@@ -192,6 +192,8 @@ template: |
       {{ toYaml .Values.global.proxy.lifecycle | indent 4 }}
     {{- end }}
     env:
+    - name: SIGN_CERT_AT_KUBERNETES_CA
```
howardjohn: this one is not needed though, we can just detect based on file, right? otherwise lgtm

lei-tang: Yes, this one is not strictly needed and the detection will be based on the existence of the file under a special directory. I have removed this one from the PR.

howardjohn (approved):

lei-tang:
@howardjohn The test noauth_installer failed with the following error:
error: timed out waiting for the condition on deployments/istio-ingressgateway.

The noauth_installer test seems to have problems since other PRs in this repo also failed at this test with the same error, e.g., the PR https://github.com/istio/installer/pull/644 also has the same error in https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_installer/644/noauth_installer/448.

Should the test noauth_installer be temporarily disabled before it is fixed?

howardjohn:
we should fix it. most likely a change in istio (possibly yours) changed the behavior? see error "unknown service istio.v1.auth.IstioCertificateService" from ingress

lei-tang:
How to find the error "unknown service istio.v1.auth.IstioCertificateService" from ingress?
I searched "IstioCertificateService" in the build log (https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_installer/645/noauth_installer/450/build-log.txt) but did not find it.

howardjohn:
on the test page click Artifacts and if you click through you can
eventually see the logs for Ingress

lei-tang:
Found it: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_installer/645/noauth_installer/450/artifacts/kind/istio-testing-control-plane/containers/istio-ingressgateway-6948d9d69f-qqvfb_istio-system_istio-proxy-479881273f5da97512fb914f1c44748e90c3b273477280ef3687b12b6786b8c3.log

I will diagnose this error.

lei-tang:
@howardjohn I created a fix for the above problem: https://github.com/istio/istio/pull/20041. Please review it.

PR #647, opened 2020-01-13
Add annotation for enableCoreDump

ckcd:
manually copy of `injection-template.yaml`.

See https://github.com/istio/istio/pull/20096


morvencao (commented):
/ok-to-test

clyang82 (approved):

PR #648, opened 2020-01-14
Configure Pilot certificate provider

lei-tang:
Configure Pilot certificate provider. Currently, two providers are supported: "kubernetes" and "citadel".


howardjohn (changes requested):
This should be done in istio/istio now

lei-tang:
Moved to istio/istio.

PR #664, opened 2020-03-03
fix mixer can't start when control plane security is disabled

zhranklin:
When istio-telemetry started with `global.controlPlaneSecurityEnabled`=false, connection to it is refused because there is no sidecar with istio-telemetry.
I notice that this bug is fixed in branch release-1.4 and master. But we are using istio 1.3 and I think there will be many other people also using 1.3. I think this should be fixed in release-1.3.


sdake:
istio/installer has been integrated into the main istio/istio tree. This repo is out of the support window. Additionally enabling control plane security flag is gone. If you wish to reopen this PR, feel free.

PR #665, opened 2020-03-03
fix: namespace isolation for ClusterRole and ClusterRoleBinding of kiali

zhranklin:
add namespace isolation for ClusterRole and ClusterRoleBinding of kiali so that multiple kiali can be installed.

### DIFF
--- a/istio-telemetry/kiali/templates/clusterrole.yaml
+++ b/istio-telemetry/kiali/templates/clusterrole.yaml
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: kiali
+  name: kiali-{{ .Release.Namespace }}
   labels:
     app: kiali
     release: {{ .Release.Name }}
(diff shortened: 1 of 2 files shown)


richardwxn (commented):
can you post PR to https://github.com/istio/istio/tree/master/manifests instead, we merge installer into istio/istio already

jmazzitelli:
This isn't how multi-tenant installations of Kiali should be handled. Installing Kiali via the Kiali Operator is how it should be done. Multiplying clusterroles/bindings may get it installed, but I have a feeling other things may go wrong (can't think of anything off hand) - but the point is, multi-tenant or even multiple copies of Kiali should be handled by running a Kiali Operator and letting it handle installing multiple copies of Kiali.

zhranklin:
@jmazzitelli It's not for multi-tenant. We use multiple deployments for such as canary deployment and debugging telemetry without influence to telemetry environment which is already used by other developers.

zhranklin:
@richardwxn OK. I've done this in <https://github.com/istio/istio/pull/21754>

But now there is a problem that users of istio 1.3 (such as me, in product environment) may still have to use the code in the release-1.3 branch for deploying istio components. Shall we update this change to the release-1.3 branch of this repository?

jmazzitelli:
> @jmazzitelli It's not for multi-tenant. We use multiple deployments for such as canary deployment and debugging telemetry without influence to telemetry environment which is already used by other developers.

I'm not saying it WON'T work but I will say that Kiali hasn't been tested or supported with multiple installations in the same cluster unless installed via the Kiali Operator.

